PR #1, opened 2021-06-25
Header component

saifulislamdev:
I'll keep in mind to have more meaningful commit messages.


public/locales/en/Header.json
```diff
@@ -0,0 +1,20 @@
+{
```
aannyyaa: Since we will not have that much content in our translation json files besides the text for the header and footer, I wonder if we should just have one translation file for each language? Let's look into the affects on performance of having multiple files. Also, then we will not have to deal with multiple name spaces

src/components/LanguageSwitcher.module.css
```diff
@@ -0,0 +1,11 @@
+/* TODO: BEMS convention */
```
aannyyaa: - I think we should avoid using css modules and instead use our themes and classNames 
- Each time we merge to dev, I think it would be best to already be using the BEM naming conventions. That way we don't have to change the classNames in PRs that are unrelated or worry about forgetting to change them

src/components/header.js
```diff
@@ -1,7 +1,129 @@
...
+    es: "spanish",
+  };
+
+  const [expanded, setExpanded] = useState(false);
+  const onClick = () => setExpanded((prvExpanded) => !prvExpanded);
+  const [isOpen1, setIsOpen1] = useState([false]);
+  const [isOpen2, setIsOpen2] = useState([false]);
+
+  const testMenuItemsOne = [
+    <Link href="#linkOne" key="one">
```
aannyyaa: I believe we should use 'to' not 'href'. Also let's look into whether we should use Link from react-uswds or from Gatsby

src/components/header.js
```diff
@@ -1,7 +1,129 @@
...
+        // isCurrent={true} // TODO: update later
+      />
+      <Menu
+        key="two"
+        items={testMenuItemsTwo}
+        isOpen={isOpen2[0]}
+        id="testDropDownTwo"
+      />
+    </React.Fragment>,
+    <a href="#three" key="three" className="usa-nav__link"> {/* TODO: doesn't like when I use NavLink instead */}
```
aannyyaa: I think we should be able to avoid anchor tags. What is the error for NavLink?

src/components/header.js
```diff
@@ -1,7 +1,129 @@
...
+      <span>{t('nav.parentOne')}</span>
+    </a>
+  ];
+
+  return (
+    <React.Fragment>
```
aannyyaa: Do we need React.Fragment here - HeaderUSWDS seems to wrap everything?

src/pages/index.js
```diff
@@ -1,13 +1,19 @@
...
+function Home() {
+  const { t, i18n } = useTranslation();
   return (
-    <div>
-      <Header />
-      <h1>Hello World!</h1>
-    </div>
+    <Suspense fallback="loading">
+      <div>
+        <Header t={t} i18n={i18n} />
```
aannyyaa: I wonder if we should use the I18nextProvider instead of passing as props: https://react.i18next.com/latest/i18nextprovider

saifulislamdev: Great point, used i18nextProvider instead.

src/components/header.js
saifulislamdev: I had something before the `HeaderUSWDS` wrap so I had to use `React.Fragment `at the time. Not necessary anymore, thanks!

src/components/header.js
saifulislamdev: Fixed now. For some reason, it showed there was a `NavLink` component in `@trussworks/react-uswds` that I could use but it kept giving me errors. Ended up switching to the `NavLink` component in `react-router-dom`, which is what I also used in the first project.

src/components/header.js
saifulislamdev: Thanks, at the time I was wondering because I realized it allowed only `href` and was trying to figure out if it would make a difference. Then, I realized I was still trying to import it from `@trussworks/react-uswds`. Also, I was going to mention that! Let's bring it up to Rapi.

src/components/LanguageSwitcher.module.css
saifulislamdev: These CSS modules are for when we need to use custom styling that is not covered in USWDS. Did you want to use a simple stylesheet instead? Also, I was trying to use the BEMS convention with CSS modules and it would not like the hyphens in class selectors when incorporating in HTML, so I figured that I would leave that comment to come back to it later.

public/locales/en/Header.json
saifulislamdev: What about the case where there are many links in the navigation bar like some government sites have?

aannyyaa: Looking at the docs (https://www.i18next.com/principles/namespaces), I think we'd want to have multiple namespaces if we were using the json files for page content, but we're using markdown for that. I think our project is small enough to just have one file for each language

src/components/LanguageSwitcher.module.css
aannyyaa: I think we'd want to add classes in a scss file for custom styling, not css

aannyyaa: It should also probably be in our styles folder (which I created in the branch I'm working on today), instead of components

public/locales/en/Header.json
saifulislamdev: Completed! `public/locales/en/translation.json` and `public/locales/es/translation.json` are created.

src/components/LanguageSwitcher.module.css
saifulislamdev: I believe Justin said a while back to use CSS files for custom classes.

aannyyaa (approved):

PR #3, opened 2021-06-28
Polishing Header component

saifulislamdev:
More meaningful commit messages :)


src/components/header.js
```diff
@@ -9,25 +9,30 @@ import {
...
   const [expanded, setExpanded] = useState(false);
   const onClick = () => setExpanded((prvExpanded) => !prvExpanded);
 
-  const [isOpen1, setIsOpen1] = useState([false]);
-  const [isOpen2, setIsOpen2] = useState([false]);
+  /* nav dropdown expansions */
+  const [isOpen1, setIsOpen1] = useState(false);
+  const [isOpen2, setIsOpen2] = useState(false);
 
+  /* first dropdown items */
```
aannyyaa, line 37: Thanks for adding these comments highlighting what these are for!

src/components/header.js
```diff
@@ -99,11 +108,13 @@ const Header = () => {
 
   return (
     <HeaderUSWDS extended={true}>
-      {/* <a class="usa-skipnav" href="#main-content">Skip to main content</a> */}
+      <a class="usa-skipnav" href="#main-content">
```
aannyyaa: I think there might be a way to do this through gatsby's Link component

src/styles/Header.css
```diff
@@ -0,0 +1,3 @@
+.search-bar {
```
aannyyaa: When we have the styling files set up, this should probably be in our scss file of class names. I think we should also be using BEM

src/components/LanguageSwitcher.js
```diff
@@ -1,7 +1,7 @@
 import React from "react";
 import { Dropdown, Label } from "@trussworks/react-uswds";
 import { useTranslation } from "react-i18next";
-import "./LanguageSwitcher.css";
+import "../styles/LanguageSwitcher.css";
```
aannyyaa, line 4: I think we want to use scss files, not css

src/components/header.js
saifulislamdev: They have something even better with `import { SkipNavLink, SkipNavContent } from "@reach/skip-nav";`. Check out the new push.

src/styles/Header.css
saifulislamdev: I did this with the intention that I was doing it in BEM, but now I realize I wasn't. Nice catch. So would it be `.header__search`?

src/components/LanguageSwitcher.js
saifulislamdev: In the last PR, I mentioned: 

> I believe Justin said a while back to use CSS files for custom classes.

 We can address this with him if you'd like.

aannyyaa: Yes, I believe we would have all our themes in the settings.scss file and then we'd have one or more scss files with class names. We want to use scss so we can access variables. I don't think we'd want to limit ourselves to a regular css file

src/components/header.js
```diff
@@ -1,5 +1,5 @@
 import React, { useState } from "react";
-import { Link, NavLink, BrowserRouter as Router } from "react-router-dom";
+import { Link, NavLink, BrowserRouter as Router } from "react-router-dom"; // TODO: use Link and NavLink from Gatsby and USWDS instead
```
aannyyaa, line 2: Might make sense to switch to gatsby Link in this PR

PR #4, opened 2021-06-29
Components in body & frontmatter formatting

aannyyaa:


src/components/graphic.js
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+import { GridContainer, Grid } from "@trussworks/react-uswds";
+
+const Graphic = () => {
+  return (
+    <GridContainer>
```
saifulislamdev: We will most likely have a `GridContainer` in the outermost of the `Body` component or `main`, so will this be necessary? You can simply create a `Grid` and leave out the `GridContainer`.

src/components/hero.js
```diff
@@ -0,0 +1,18 @@
+import React from "react";
+import { GridContainer, Grid, Button } from "@trussworks/react-uswds";
+
+const Hero = ({ heroCalloutHeading, heroCalloutText, callToActionButton }) => {
+  return (
+      <GridContainer>
```
saifulislamdev: Same as previous comment – will `GridContainer` be necessary?

src/components/index.js
```diff
@@ -0,0 +1,5 @@
+export {default as Graphic} from './graphic'
```
saifulislamdev, line 1: Is this the same as `export Graphic from './graphic'`?

src/components/section.js
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+import { GridContainer, Grid } from "@trussworks/react-uswds";
+
+const Section = () => {
+  return (
+    <GridContainer>
```
saifulislamdev: Same as previously mentioned comments – is `GridContainer` necessary here?

src/components/tagline.js
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+import { GridContainer, Grid } from "@trussworks/react-uswds";
+
+const Tagline = () => {
+  return (
+    <GridContainer>
```
saifulislamdev: Same as previously mentioned comments – is `GridContainer` necessary here?

src/pages/{MarkdownRemark.frontmatter__slug}.js
```diff
@@ -1,17 +1,21 @@
...
   const { frontmatter } = markdownRemark;
   return (
-    <div>
-      <div>
-        <h1>{frontmatter.heroCalloutHeading}</h1>
-        <p>{frontmatter.heroCalloutText}</p>
-        <button>{frontmatter.callToActionButton}</button>
-      </div>
-    </div>
+    <main>
```
saifulislamdev: It would probably be better to have the `GridContainer` here instead.

src/components/graphic.js
```diff
@@ -1,11 +1,15 @@
 import React from "react";
 import { GridContainer, Grid } from "@trussworks/react-uswds";
 
-const Graphic = () => {
+const Graphic = ({ images, graphicHeadings }) => {
   return (
     <GridContainer>
```
saifulislamdev: Same as previously mentioned comments – is `GridContainer` necessary here?

src/components/graphic.js
```diff
@@ -1,11 +1,15 @@
...
-      <Grid col>
-        <Grid></Grid>
+      <Grid row>
+        {images.map((image) => (
+          <Grid col={3}>
+            <p>{image}</p>
```
saifulislamdev: What is `image`? If it's actually an image, do you want to put it in between a `<p>` tag? If it's text and not an image, I would suggest using a more descriptive variable name (i.e. `imageText`).

src/components/section.js
```diff
@@ -1,11 +1,14 @@
...
 import { GridContainer, Grid } from "@trussworks/react-uswds";
 
-const Section = () => {
+const Section = ({ sectionHeading, sectionText }) => {
   return (
     <GridContainer>
-      <Grid col>
-        <Grid></Grid>
+      <Grid row>
+        <Grid>
```
saifulislamdev: Is it necessary to have another `<Grid>` here? Can we get away with using `<Grid row>` only?

src/components/tagline.js
```diff
@@ -1,11 +1,14 @@
...
 import { GridContainer, Grid } from "@trussworks/react-uswds";
 
-const Tagline = () => {
+const Tagline = ({taglineHeading, taglineText}) => {
   return (
     <GridContainer>
-      <Grid col>
-        <Grid></Grid>
+      <Grid row>
+        <Grid>
```
saifulislamdev: Is it necessary to have another `<Grid>` here? Can we get away with using `<Grid row>` only?

src/components/tagline.js
```diff
@@ -1,11 +1,14 @@
 import React from "react";
 import { GridContainer, Grid } from "@trussworks/react-uswds";
 
-const Tagline = () => {
+const Tagline = ({taglineHeading, taglineText}) => {
```
saifulislamdev: I'm assuming more props will be added that make it different from the `Section` component, but right now, they seem like the same component.

src/components/graphic.js
aannyyaa: I think you're not commenting on the most up to date version. I was just doing this to see if the content was coming through. It should be an img tag now

src/components/graphic.js
```diff
@@ -1,15 +1,11 @@
...
+const Graphic = ({}) => {
   return (
     <GridContainer>
       <Grid row>
-        {images.map((image) => (
-          <Grid col={3}>
-            <p>{image}</p>
-          </Grid>
-        ))}
+        <Grid></Grid>
```
saifulislamdev: Is it necessary to have another <Grid> here? Can we get away with using <Grid row> only?

aannyyaa: You're commenting on outdated files somehow

src/components/section.js
```diff
@@ -1,13 +1,14 @@
 import React from "react";
-import { GridContainer, Grid } from "@trussworks/react-uswds";
+import { GridContainer, Grid, Button } from "@trussworks/react-uswds";
 
-const Section = ({ sectionHeading, sectionText }) => {
+const Section = ({ sectionHeading, sectionText, callToActionButton }) => {
```
saifulislamdev: For readability, it may be better to call this `callToActionButtonText`.

src/components/tagline.js
aannyyaa: Section and tagline are different if you look at the uswds landing page. They'll have slightly different layout

src/components/section.js
```diff
@@ -1,13 +1,13 @@
...
 
-const Section = ({ sectionHeading, sectionText, callToActionButton }) => {
+const Section = ({ section, callToActionButton }) => {
   return (
     <GridContainer>
       <Grid row>
         <Grid>
-          <h2>{sectionHeading}</h2>
-          <p>{sectionText}</p>
+          <h2>{section.sectionHeading}</h2>
```
saifulislamdev: `section.sectionHeading` --> `section.heading`?

src/components/section.js
```diff
@@ -1,13 +1,13 @@
...
+          <p>{section.sectionText}</p>
```
saifulislamdev: `section.sectionText` --> `section.text`?

src/components/tagline.js
```diff
@@ -1,13 +1,13 @@
...
 
-const Tagline = ({taglineHeading, taglineText}) => {
+const Tagline = ({tagline}) => {
   return (
     <GridContainer>
       <Grid row>
         <Grid>
-          <h2>{taglineHeading}</h2>
-          <p>{taglineText}</p>
+          <h2>{tagline.taglineHeading}</h2>
```
saifulislamdev: `tagline.taglineHeading` --> `tagline.heading`?

src/components/tagline.js
```diff
@@ -1,13 +1,13 @@
...
+          <p>{tagline.taglineText}</p>
```
saifulislamdev: `tagline.taglineText` --> `tagline.text`?

src/components/section.js
aannyyaa: I actually made that change yesterday. Again, you're looking at outdated code

src/components/graphic.js
```diff
@@ -7,7 +7,10 @@ const Graphic = ({ graphics }) => {
       <Grid row>
         {graphics.map((graphic) => {
           return (
-            <Grid mobile={{ col: "fill" }} tablet={{ col: 6 }}>
+            <Grid
```
saifulislamdev: I feel like you left out the `desktop` attribute for a certain reason.

src/pages/{MarkdownRemark.frontmatter__slug}.js
```diff
@@ -1,6 +1,8 @@
 import React from "react";
 import { graphql } from "gatsby";
 import { Graphic, Hero, Section, Tagline } from "../components/";
+import "@trussworks/react-uswds/lib/uswds.css";
```
saifulislamdev: I imported these in `index.js`.

src/components/index.js
aannyyaa: I don't think that syntax works. Also I'm exporting all the components like that from components/index.js so they can easily be imported like this: import { Graphic, Hero, Section, Tagline } from "../components/";

saifulislamdev: Never mind, I read this wrong. Great job.

src/components/graphic.js
```diff
@@ -1,24 +1,20 @@
...
           return (
-            <Grid
-              mobile={{ col: "fill" }}
-              tablet={{ col: 6 }}
-            >
-              <img src={graphic.image} alt={graphic.imageDescription} />
-              <h2>{graphic.heading}</h2>
-              <p>{graphic.text}</p>
+            <Grid mobile={{ col: "fill" }} tablet={{ col: 6 }}>
+              {<img src={graphic.image} alt={graphic.imageDescription} />}
```
saifulislamdev: You don't need surrounding curly braces here.

src/templates/landing-page.js
```diff
@@ -0,0 +1,69 @@
+import React from "react";
+import { graphql } from "gatsby";
+import { GridContainer } from "@trussworks/react-uswds";
+import { Graphic, Header, Hero, Section, Tagline } from "../components";
+import "@trussworks/react-uswds/lib/uswds.css";
```
saifulislamdev, line 5: Are these imported in `index.js` as well?

saifulislamdev (changes requested):
Check out additional comments.

src/components/graphic.js
aannyyaa: Forgot to add my graphic.image && logic there

src/templates/landing-page.js
aannyyaa: I believe we won't use index.js going forward and will use landing-page instead, so I imported here. Also, if we were to have multiple pages, we might need to import those multiple times, since the file structure is different than with create-react-app and we don't import all the other pages into index.js eventually. But we should look into that!

src/pages/{MarkdownRemark.frontmatter__slug}.js
saifulislamdev: Where did your GridContainer go?

saifulislamdev (approved):
Good.

PR #5, opened 2021-06-29
Custom Link component

saifulislamdev:


src/components/Link.js
```diff
@@ -0,0 +1,65 @@
...
+
+  if (variant === "nav") {
+    // Navigation link
+    return (
+      // TODO: there's no special attribute/component that Gatsby has for navigation, so should Gatsby links be used here?
+      <USWDSLink href={to} variant="nav" {...other}>
+        {children}
+      </USWDSLink>
+    );
+  } else if (internal) {
```
aannyyaa: A file download is an internal link?

src/components/Link.js
```diff
@@ -0,0 +1,65 @@
...
+          {/* Show file type and size for links to non-HTML content */}
+          {`[${fileType.toUpperCase()}, ${size}]`}
+        </USWDSLink>
+      );
+    }
+    return (
+      // Use Gatsby Link for internal links
+      <GatsbyLink
+        to={to}
+        className="usa-link" // TODO: would this technically be considered hard-coding? What if USWDS decides to change the class name to something else?
```
aannyyaa, line 54: Good point. But for the most part I think we can rely on the class name to remain constant, while the exact styles of that class name could change

src/components/Link.js
```diff
@@ -0,0 +1,65 @@
...
+        <USWDSLink href={to} {...other}>
+          {/* TODO: Download attribute available but not sure what it does or is for */}
+          {/* TODO: Should file download links be wrapped with Gatsby as well? Not sure if there's preprocessing there. */}
+          {children}
+          {/* Show file type and size for links to non-HTML content */}
+          {`[${fileType.toUpperCase()}, ${size}]`}
+        </USWDSLink>
+      );
+    }
+    return (
```
aannyyaa, line 50: It's kind of hard for me to tell where the if/else statement ends when looking at this in github, but should there be an else on line 44? Also, shouldn't the external links (starting on 57) be included in the if/else logic as well?

src/components/Link.js
```diff
@@ -0,0 +1,65 @@
...
+  partiallyActive,
+  variant,
+  fileType,
+  size,
+  ...other
+}) => {
+  // Assumes that any internal link will start with exactly one slash, and that anything else is external.
+  const internal = /^\/(?!\/)/.test(to);
+
+  // Assumed file download link structure
```
aannyyaa, line 20: What is the assumed file download link structure?

src/components/Link.js
```diff
@@ -0,0 +1,65 @@
...
+  to,
+  activeClassName,
+  partiallyActive,
+  variant,
+  fileType,
+  size,
+  ...other
+}) => {
+  // Assumes that any internal link will start with exactly one slash, and that anything else is external.
+  const internal = /^\/(?!\/)/.test(to);
```
aannyyaa, line 18: You could also check for different domain name?

src/components/Link.js
```diff
@@ -0,0 +1,65 @@
...
+
+  // Assumed file download link structure
+  const file = /\.[0-9a-z]+$/i.test(to);
+
+  if (variant === "nav") {
+    // Navigation link
+    return (
+      // TODO: there's no special attribute/component that Gatsby has for navigation, so should Gatsby links be used here?
```
aannyyaa: Is a nav link functionally the same as a normal link, but stylistically different, as it is for react router dom links (https://reactrouter.com/web/api/NavLink)? If so, then I think we should use gatsby for internal nav links

src/components/Link.js
saifulislamdev: This is how it is under 'Reminder: use <Link> only for internal links!' and 'File Downloads'. It kind of makes sense, too; file downloads will be for files in `public/` or `src/` folders.

https://www.gatsbyjs.com/docs/reference/built-in-components/gatsby-link/

src/components/Link.js
saifulislamdev: There can be an `else` on line 44. But there doesn't necessarily have to be as it works the same with and without it. It looks cleaner that way also. Hence, the external links starting on 57 are included in the if/else logic in a different way.

For example, this works the same without an `else`.
```javascript
function oddOrEven(n) {
  if (n % 2) return 'odd';
  return 'even';
  // else return 'even';
}
```

src/components/Link.js
saifulislamdev: Good question! I took it directly from the Gatsby documentation. But from my understanding, the file download link structure satisfies with a period in the beginning, followed by 1 or more numbers or letters, and then the end of the line is reached. The `/i` at the end ignores the case of the string (case-insensitive).

src/components/Link.js
saifulislamdev: By satisfying this regex, it satisfies that it's in the same domain name. It checks to see if the entered link is a route. If the entered link is a route, we can assume it's still on the same domain.

src/components/Link.js
saifulislamdev: A nav link can be the same as a normal link with extra classes added. I use Gatsby for internal nav links and USWDS for external nav links in the most recent push.

src/components/Link.js
```diff
@@ -0,0 +1,78 @@
...
+        {children}
+      </GatsbyLink>
+    );
+  }
+  // external
+  else {
+    return (
+      <USWDSLink
+        href={to}
+        variant={variant !== "nav" && "external" /* if not nav link, set it to be regular external link */}
```
aannyyaa: With this syntax we only set variant if it is not nav, right? Should we use a ternary operator instead of && syntax? Same with the syntax on the following line, if it's not nav we still want class name to be ""usa-link usa-link--external" right?

src/components/Link.js
```diff
@@ -0,0 +1,78 @@
...
+          className="usa-link usa-link--nav"
+          activeClassName={activeClassName}
+          partiallyActive={partiallyActive}
+          {...other}
+        >
+          {children}
+        </GatsbyLink>
+      );
+    }
+    if (file) {
```
aannyyaa: This is just coding style / preference, but line 38 could be "else if"

src/components/Link.js
saifulislamdev: Yes, we only set the variant if it is not nav. It works the same and looks cleaner this way without the ternary operator.

As for the following line, if it's not nav, that means it's only an external link. But if it's an external link, we don't have to add any class names as it's handled by the variant attribute since it's set to external in the previous line. Although, I should have gotten rid of the `usa-link` class because it's already in there since we are using the USWDSLink component. I'll get rid of it and push.

src/components/Link.js
saifulislamdev: You know what, looking at it now, it makes sense to include that for understandability. Push made.

PR #7, opened 2021-07-01
Layout refactor

aannyyaa:


package.json
```diff
@@ -4,6 +4,7 @@
   "description": "This is a proposed template repository for applications for future developments. This will help get you set up as you build your application.",
   "main": "index.js",
   "dependencies": {
+    "@reach/skip-nav": "^0.15.2",
```
saifulislamdev, line 7: I think I installed this dependency as well, oh well.

src/components/Layout.js
```diff
@@ -1,6 +1,5 @@
 import React from "react";
-import Header from "./Header";
-import Footer from "./Footer";
+import { Footer, Header } from ".";
```
saifulislamdev, line 2: I noticed you import components alphabetically, which is what I also usually do. But here, there are only two components imported, so should it instead be imported in the order it appears in the DOM? As in:
`import { Header, Footer } from ".";`

Or do we always import them alphabetically?

src/components/index.js
```diff
@@ -1,5 +1,10 @@
-export {default as Graphic} from './graphic'
-export {default as Header} from './header'
-export {default as Hero} from './hero'
-export {default as Section} from './section'
-export {default as Tagline} from './tagline'
\ No newline at end of file
+export { default as Graphic } from "./graphic";
```
saifulislamdev, line 1: Nice ordering.

src/templates/landing-page.js
```diff
@@ -1,32 +1,45 @@
...
-              buttons={frontmatter.buttons}
-            />
-          )}
-        </GridContainer>
-      </main>
-    </div>
+    <Suspense fallback="loading">
+      <I18nextProvider i18n={i18next}>
+        <Layout>
+          <main>
```
saifulislamdev, line 23: See if any component you use introduces `<main>` in the DOM already, you MIGHT be repeating it.

src/components/Layout.js
aannyyaa: Good point. Maybe stick with alphabetical since the order in the DOM could change? What do you think?

package.json
aannyyaa: It wasn't installed on the branch I was working on, so needed to install here otherwise I got errors. Should be fine installing twice :)

src/templates/landing-page.js
aannyyaa: Oh good point - I hadn't thought of that! Will look

PR #8, opened 2021-07-01
Footer component

saifulislamdev:


src/components/Link.js
```diff
@@ -1,72 +1,64 @@
...
 
-  if (internal) {
+  if (isInternal) {
     if (variant === "nav") {
       // Navigation link
       return (
         <GatsbyLink
           to={to}
-          className="usa-link usa-link--nav"
+          className={classNames("usa-link", "usa-link--nav", className)} // quoted classes are always included while className is included if it has a truthy value
```
aannyyaa, line 23: I didn't know about the classnames package - it's cool!

public/locales/en/translation.json
```diff
@@ -13,8 +14,34 @@
     },
     "parentOne": "Parent link"
   },
+  "secondaryLink": [],
```
aannyyaa, line 17: Should this array include the link names on lines 19 and 20?

public/locales/en/translation.json
```diff
@@ -1,6 +1,7 @@
 {
   "title": "Project Title",
   "nav": {
+    "dropdown": [],
```
aannyyaa, line 4: Did you mean to have an empty array here or should it have the dropdownOne info in it

src/components/Footer.js
```diff
@@ -0,0 +1,115 @@
...
+  const primaryLinks = ["#", "#", "#", "#"];
+
+  return (
+    <FooterUSWDS
+      size="medium"
+      // returnToTop={returnToTop}
+      primary={
+        <FooterNav
+          size="medium"
+          links={Array.from({ length: 4 }, (_, i) => {
```
aannyyaa: Can we not hardcode the array length here?

src/components/Footer.js
```diff
@@ -0,0 +1,115 @@
...
+              />
+            }
+            heading={
+              <p className="usa-footer__logo-heading">
+                {t("footer.secondary.heading.logo")}
+              </p>
+            }
+          />
+          <Grid className="usa-footer__contact-links" mobileLg={{ col: 6 }}>
+            <SocialLinks
```
aannyyaa, line 67: Instead of hardcoding in each of these social media links, socialLinks can be an array in the translation file and you can map through the array here. I guess you'd need to do some string interpolation in the classnames

src/components/Link.js
```diff
@@ -1,72 +1,64 @@
...
-  else {
+  } else {
+    // external
     return (
       <USWDSLink
         href={to}
-        variant={variant !== "nav" ? "external" : undefined /* if not nav link, set it to be regular external link */} 
-        className={variant === "nav" ? "usa-link--external usa-link--nav" : "" /* if nav link, set it to be both external and link */}
+        variant={
+          variant !== "nav"
```
aannyyaa: Change back to && syntax instead of having a ternary operator with "undefined"

public/locales/en/translation.json
saifulislamdev: The empty array is there on purpose, it was a placeholder. The array is now filled in the next PR.

public/locales/en/translation.json
saifulislamdev: The empty array is there on purpose, it was a placeholder. The array is now filled in the next PR.

src/components/Footer.js
saifulislamdev: Not hardcoded anymore; check the newest commit.

src/components/Footer.js
saifulislamdev: So if I do that, would class names like `usa-social-link--facebook` be in the translation.json file? Where would be the best place to keep classes like these?

src/components/Link.js
saifulislamdev: Hmm, I thought this wouldn't work, but it works fine. Check it out in the newest commit.

src/components/Footer.js
aannyyaa: No, I don't think you need to store the whole class name. I was thinking when you map through the array of social media links it would be something along these lines, with each element being a `socialLink`:

```
<Link
          key={socialLink.name}
           className=`usa-social-link usa-social-link--${socialLink.name}`
            to="#"
   >
```

saifulislamdev: Hard coding is gone, thanks for the helpful code. Check out the newest commit.

PR #9, opened 2021-07-01
Refactoring code in translation files

saifulislamdev:
Next steps:

1. Go over comments in #8 
1. Go over comments in #9  
1. Go over comments in #10 
1. Deploy with Azure
1. Personalized Dropdown component for nav


src/components/header.js
```diff
@@ -38,46 +38,63 @@ const Header = () => {
...
+  //       </Link>
+  //     );
+  //   });
+
+  /* TODO: try to dynamically create */
   const testMenuItemsOne = [
-    <Link to="#linkOne" key="one">
-      {t("nav.dropdownOne.simpleLinkOne")}
+    <Link to="linkOne" key="one">
+      {t("header.nav.dropdown.0.simpleLink.0")}
```
aannyyaa: I feel like this syntax is not the most readable and also wouldn't be easy to update if more links were added in the dropdown array. Can you map through the dropdown array to create the menu items, instead of hardcoding the number of menu items?

saifulislamdev: Yes, that's the goal. But I'm having the same trouble dynamically creating the links as I am on line 42. It's something I may take into account next week.

saifulislamdev: Mapped through the `dropdowns` and `parentLinks` arrays in the translation files to create parent links and the menu items for each dropdown in the navigation. Further refactoring will be done by introducing a `NavDropDown` component in the next PR.

src/components/header.js
```diff
@@ -39,40 +39,72 @@ function Header({ slug }) {
...
-      <Link to={navDropdownLinks1[i]} key={i}>
-        {navDropdownLinkLabels1[i]}
-      </Link>
-    );
-  });
+  const navDropdownLength1 =
+    navDropdownLinks1.length > navDropdownLinkLabels1.length
+      ? navDropdownLinkLabels1.length
+      : navDropdownLinks1.length;
+  const navDropdownItems1 = Array.from(
```
aannyyaa: Can you map through navDropdownLinks instead of using this syntax? If you use map then you also won't need to compare the array lengths above

src/components/header.js
```diff
@@ -39,40 +39,72 @@ function Header({ slug }) {
   const [isOpen2, setIsOpen2] = useState(false);
 
   /* first dropdown items */
-  const navDropdownLinks1 = ['link-one', 'link-two']
+  const navDropdownLinks1 = ["link-one", "link-two"];
```
aannyyaa: Why are the paths hard coded here? I think they should be in the translation files

src/components/header.js
```diff
@@ -111,10 +143,7 @@ function Header({ slug }) {
         id="navDropDownTwo"
       />
     </React.Fragment>,
-    <Link variant="nav" to="/three" key="3">
-      <span>{t("header.nav.parentLinks.0")}</span>
-    </Link>,
-  ];
+  ].concat(parentLinkItems);
```
aannyyaa, line 165: How is the concat functioning here / what is it's purpose

aannyyaa (changes requested):
Can you tell me more about why you use `Array.from`? I think `map` might be simpler

src/components/header.js
saifulislamdev: I compare the array lengths to take the shorter of the two in case the user forgets to add a label in the translation file or add a link to the array. Mapping through the navDropdownLinks can also be done, but this syntax is easier to understand if taking the shorter of the two. If you look at the `parentLinks` mapping (which you want me to do), it's more confusing to understand. Let me know which one you think I should stick to – `Array.from` or `.map`.

src/components/header.js
saifulislamdev: There will be duplication if put in the translation files because regardless of what language they're in they will be directed to the same page.

src/components/header.js
saifulislamdev: The purpose is to add the parent links to the navigation bar. The parent links are stored in their own array to account for when the user wants to have more than 1 parent link.

src/components/header.js
aannyyaa: Okay. Maybe they should be in the constants folder? Maybe all of the hardcoded links can be in one file, so they can all be updated at once and won't get lost in the code of a component

src/components/header.js
aannyyaa: I personally think map is easier to understand. 

If the user does forget to add a label or file, then the labels and files might be mismatched, if we proceed with looping through the shorter array. Maybe we could show an error if the lengths don't match?

It is definitely great that you're thinking about cases of user error though!

src/components/header.js
```diff
@@ -39,40 +39,72 @@ function Header({ slug }) {
...
+    (_, i) => {
+      return (
+        <Link to={secondaryLinks[i]} key={i}>
+          {secondaryLinkLabels[i]}
+        </Link>
+      );
+    }
+  );
 
   const navBarItems = [
```
aannyyaa, line 128: Here we're still hard coding the number of navBarItems. I think we should be able to map through an array to create however many NavDropDownButtons are needed

src/components/header.js
saifulislamdev: How would we show an error if the lengths don't match? I think what we're doing right now is good enough. I'll let you know once the mapping is complete.

public/locales/en/translation.json
```diff
@@ -1,20 +1,19 @@
 {
-  "title": "Project Title",
-  "nav": {
-    "dropdownOne": {
-      "label": "Nav Label 1",
-      "simpleLinkOne": "Simple link one",
-      "simpleLinkTwo": "Simple link two"
+  "header": {
```
aannyyaa, line 2: This translation file is still in `public`. I wonder if it would make more sense at this point, since this PR is so old, to create a new PR off of dev once we merge language switcher. Then we can update all 13 translation files, which are now in `src`.  We can also copy and paste your header and footer changes into the new branch?

src/components/header.js
aannyyaa: If they missed the first link, then all of the link / name pairs would be off. This is actually a good reason to have the links and the labels stored in the same place in an array of objects maybe?

aannyyaa: If we wanted to show an error we could do something like [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/throw). But you're right that it's probably not needed now

saifulislamdev: Mapping is complete.

src/components/header.js
saifulislamdev: Excellent thought and reasoning. Links added to `/src/constants/links.js`.

src/components/header.js
```diff
@@ -39,7 +41,7 @@ function Header({ slug }) {
   const [isOpen2, setIsOpen2] = useState(false);
 
   /* first dropdown items */
-  const navDropdownLinks1 = ["link-one", "link-two"];
+  const navDropdownLinks1 = links.navDropdown1;
```
aannyyaa: Thanks for moving the links to the constants file! 

Also, can you use an array and mapping to create the first and second dropdown items, instead of hard coding? A user might want 3 nav drop downs for example, and shouldn't have to edit the code, only the the translation.json file

src/components/header.js
saifulislamdev: This will be something I focus on when creating the `NavDropDown` component in an upcoming PR.

public/locales/en/translation.json
saifulislamdev: Sounds good.

src/components/header.js
saifulislamdev: Links in the constants file are refactored better (use arrays instead of incrementing variable names). I used mapping to create the dropdown items, so now a user can use 3 nav dropdowns, and it will be stored accordingly. I will work on dynamically creating the `NavDropDownButton`s in another PR.

src/components/header.js
```diff
@@ -40,41 +40,39 @@ function Header({ slug }) {
...
+      : translationFileLength; // take shorter length in case there is missing dropdowns in `../constants/links.js` (constants file) or `../locales/` (translation files)
+  dropdownLinks.map((_, i) => {
+    if (i < length) {
+      const navDropdownLinks = dropdownLinks[i];
+      const navDropdownLinkLabels = t("header.nav.dropdowns")[i]["simpleLinks"];
+      const navDropdownLength =
+        navDropdownLinks.length > navDropdownLinkLabels.length
+          ? navDropdownLinkLabels.length
+          : navDropdownLinks.length; // take shorter length in case there is a missing link in `../constants/links.js` (constants file) or label in  `../locales/` (translation files)
+      dropdowns.push(
```
aannyyaa, line 85: I think pushing into the `dropdowns` array like this might not be best practice, since you don't know which index link would be missing. If there are 10 links, and the 5th is missing, then the rest of the labels would be off by one.

This is why I was advocating for placing the link and the link label in an object in the translation.json file. Something along these lines:

**EDIT**: perhaps it can be in `constants/links.js` like this:

```
dropdowns: [
{label: t("label1"), link: "/path1"},
{label: t("label2"), link: "/path2"}
]
```

src/components/header.js
```diff
@@ -105,7 +103,7 @@ function Header({ slug }) {
       />
       <Menu
         key="one"
-        items={navDropdownItems1}
+        items={dropdowns[0]}
```
aannyyaa, line 142: I think nav bar items can be created by mapping through dropdowns, then concatting parentLinkItems as you already do. Here it's still hardcoded to 2 nav dropdown buttons

aannyyaa: Sorry saw that you said you'd work on this in another PR

src/components/header.js
saifulislamdev: > If they missed the first link, then all of the link / name pairs would be off. This is actually a good reason to have the links and the labels stored in the same place in an array of objects maybe?

Are you thinking about doing this in the translation files? If so, what if the user makes a mistake when typing in the link in a random translation file while the others are correct? This won't stand out to the user and may be dismissed, while the current approach makes sure the links are intact. If they cannot see one of their links in a different language, they will know they messed up with the labeling, and their mistake will be apparent, but making sure the links are correct across translation files is a hassle. It won't be apparent to the user unless they dig deep.

If you're thinking about doing this in the `link.js` constant file, this might throw the user off a bit since they will now have to translate in 3 different places (translation files, markdown files, and constant file).

| Method | Pros | Cons |
| -------- |  -------- |  --------  |
| Labels in Translation File + Links in Constant File | <ul><li>No data duplication</li><li>Possible errors are more apparent to the user</li></ul>  | <ul><li>Some room for user error</li></ul> |
| Labels + Links in Translation File | | <ul><li>Data duplication</li><li>Links can be off and won't be apparent to the user</li><li>More room for user error</li></ul> |
| Labels + Links in Constant File | <ul><li>No data duplication</li><li>Possible errors are more apparent to the user </li></ul>| <ul><li>Some room for user error</li><li>Translating in 3 different places (i18n translation, markdown, constant files)</li></ul> |

Let me know what you think.

aannyyaa: What if you use i18next in the `link.js` constant file for the labels

src/components/header.js
aannyyaa: Great!!

src/components/header.js
saifulislamdev: Check out my previous comment.

aannyyaa: I just edited this example - using i18next in constants  and storing the link and label translation in an array of objects?

src/components/header.js
saifulislamdev: So you mean using `t()` in `link.js` to get the translations? It's a constant, and i18next will change the labels on every language change. Should we still do that?

aannyyaa: Oh do you mean is it bad practice to use a variable in a constants file? I guess the link is the actual constant that we're worried about, but we still want the label to change for each language? If you want to address the potential mismatch I think this makes sense. 

Could we pull the links into the translation.json file from constants/links? I'm not sure if you can import things into a json file

saifulislamdev: Whatever is in a constant file isn't mean to change at all, and by using i18next in the constant file, we're going against this. Not only this but will what you're thinking of work? Also, your idea is trying to help the user make sure they don't make any mistakes in the labeling, but they're going to be working in the translation files anyways, so is this really that necessary?

I don't think you can import things into a JSON file; Google is super unhelpful, at least.

aannyyaa: Okay. Mostly, I am still just unsure of going ahead and pushing the labels/links in even if the array lengths don't match

saifulislamdev: Console warning messages added!

PR #10, opened 2021-07-01
Change content language & preliminary multilingual routing

aannyyaa:


src/components/LanguageSwitcher.js
```diff
@@ -16,14 +17,18 @@ function LanguageSwitcher() {
     i18n.changeLanguage(target.value);
   };
 
+  useEffect(() => {
```
saifulislamdev, line 20: As a note, we will want to check for edge cases in the future. Especially, if this works in nested routes. No push needed.

src/components/header.js
```diff
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React, { useContext, useState } from "react";
```
saifulislamdev: I see you import this but don't use it. Is this for the future?

src/templates/landing-page.js
```diff
@@ -1,25 +1,26 @@
...
+
   return (
     <Suspense fallback="loading">
-      <I18nextProvider i18n={i18next}>
-        <Layout>
+      <I18nextProvider i18n={i18n}>
+        <Helmet
+          title={t("title")}
+          htmlAttributes={{ lang: i18n.language }}
+          defer={false}
```
saifulislamdev: What does the `defer` attribute do?

src/components/header.js
aannyyaa: Yeah I was trying to use it in order to access i18n from the I18nextProvider. I need to look into it more!

src/templates/landing-page.js
aannyyaa: Oh I actually don't need that - I'll delete it and the useContext hook. I realized defer is just suggested if you're developing using gatsby-plugin-offline

src/components/LanguageSwitcher.js
aannyyaa: Yeah this is definitely preliminary just so I could see the language content change. We can look into best practices for multilingual routing in gatsby

PR #11, opened 2021-07-02
Update file structure

aannyyaa:
- Move locales to src
- Update i18n config
- add public to .gitignore


src/components/i18n.js
```diff
@@ -8,10 +8,18 @@ i18n
...
   .init({
-    // ns: ["translations"],
-    // defaultNS: "translations",
+    ns: ["translations"],
+    defaultNS: "translations",
     debug: true,
     fallbackLng: "en",
+    resources: {
+      en: {
+        translations: require("../locales/en/translation.json"),
```
saifulislamdev, line 17: This seems to be duplicating line 9 in `gatsby-node.js`. Either delete `onPostBuild()` in `gatsby-node.js` (preferred) or delete lines 15-22 in `i18n.js`.

saifulislamdev (approved):

PR #13, opened 2021-07-06
Sass settings and themes

aannyyaa:


src/templates/landing-page.js
```diff
@@ -8,6 +8,8 @@ import { Graphic, Hero, i18n, Layout, Section, Tagline } from "../components/ind
 import "@trussworks/react-uswds/lib/uswds.css";
 import "@trussworks/react-uswds/lib/index.css";
 
+import '../styles/index.scss'
```
saifulislamdev, line 11: Later we will either want to rename this file to landing-page.scss or have this in the index.js file and not rename it (the goal).

saifulislamdev (approved):

PR #15, opened 2021-07-07
Theming

aannyyaa:


src/components/graphic.js
```diff
@@ -1,35 +1,41 @@
 import React from "react";
-import { Grid } from "@trussworks/react-uswds";
+import { Grid, GridContainer } from "@trussworks/react-uswds";
```
aannyyaa, line 2: I added GridContainer back in because it creates spacing/formatting that aligns with the USWDS template. The code example in the trustworks template also uses multiple GridContainer components.

src/components/header.js
```diff
@@ -135,7 +136,6 @@ function Header({ slug }) {
       >
         <Search
           id="search"
-          className="header__search"
```
aannyyaa, line 138: When there was a min-width, the search bar was too wide to be viewed in the menu navigation panel that opens for tablet/mobile views

src/components/language-switcher.js
```diff
@@ -23,24 +22,30 @@ function LanguageSwitcher({ slug }) {
...
-          return (
-            <option name="lng" value={lng.code} key={lng.code}>
-              {lng.label}
-            </option>
-          );
-        })}
-      </Dropdown>
-      {/* <Button type="submit">Translate</Button> */}
-      {/* TODO: good to include a submit button (for accessibility reasons)*/}
+      <Grid row gap>
```
aannyyaa, line 25: This is still in progress - trying to use grid instead of flex to position the dropdown

src/components/language-selector.js
```diff
@@ -0,0 +1,77 @@
+import React, { useEffect, useState } from "react";
+import { navigate } from "gatsby";
+import { Grid, Link, Menu, NavDropDownButton } from "@trussworks/react-uswds";
+import globe from "../images/logos/globe.png";
+
+function LanguageSelector() {
```
aannyyaa: I created language selector in another file but it will serve the same purpose as language switcher. I didn't want to interfere with our ability to select a language which working on this

src/components/language-switcher.js
aannyyaa: This component will end up being in the banner - so disregard any styling changes here for now

src/markdown-pages/landing-es.md
```diff
@@ -13,19 +13,19 @@ tagline:
   heading: es A tagline highlights your approach
   text: es The tagline should inspire confidence and interest, focusing on the value that your overall approach offers to your audience. Use a heading typeface and keep your tagline to just a few words, and don’t confuse or mystify. Use the right side of the grid to explain the tagline a bit more. What are your goals? How do you do your work? Write in the present tense, and stay brief here. People who are interested can find details on internal pages.
 graphics:
-  - image: https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.app.cloud.gov/preview/uswds/uswds/develop/dist/img/circle-124.png
+  - image: http://placehold.it/150g
```
aannyyaa, line 16: Used gray placeholder images so the previous blue (which was uswds primary) did not cause any confusion when looking at the design

package.json
```diff
@@ -13,22 +18,19 @@
...
-    "@trussworks/react-uswds": "^2.0.0",
-    "classnames": "^2.3.1",
-    "i18next": "^20.3.2",
-    "i18next-browser-languagedetector": "^6.1.2",
-    "i18next-http-backend": "^1.2.6",
-    "i18next-xhr-backend": "^3.2.2",
     "react-i18next": "^11.11.0",
-    "react-router-dom": "^5.2.0"
+    "react-router-dom": "^5.2.0",
+    "uswds": "^2.10.0"
```
saifulislamdev, line 34: Was downgrading `uswds` the fix?

src/components/banner.js
```diff
@@ -0,0 +1,34 @@
...
+} from "@trussworks/react-uswds";
+import { LanguageSelector } from ".";
+import nycLogo from "../images/logos/nyc_logo.png";
+
+function Banner({ children }) {
+  return (
+    <div className="usa-section--dark">
+      <GridContainer>
+        <Grid row  className="banner__logo">
+          <Grid row gap className="banner__logo">
```
saifulislamdev: Does the class name have to be repeated? It's already on line 14.

src/components/footer.js
```diff
@@ -52,16 +52,11 @@ function Footer() {
             size="medium"
             image={
               <img
-                className="usa-footer__logo-img"
```
saifulislamdev, line 55: How come the class was removed?

src/components/header.js
saifulislamdev: Sure. If we want to use the search bar in the future, we have to make sure to give it a minimum space to fit ~27 characters (don't remember the exact number).

package.json
aannyyaa: Not sure which package was the problem since I downgraded multiple at once - but it's possible that it was the uswds version in particular!

src/components/footer.js
aannyyaa: It made the logo too small - I ended up creating a new className today

src/components/language-selector.js
```diff
@@ -0,0 +1,77 @@
...
+    <Link to="/en/landing" key="en">
+      Русский
+    </Link>,
+    <Link to="/en/landing" key="en">
+      বাংলা
+    </Link>,
+    <Link to="/en/landing" key="en">
+      한국어
+    </Link>,
+    <Link to="/en/landing" key="en">
```
saifulislamdev: Don't need these many languages anymore.

src/components/language-selector.js
```diff
@@ -0,0 +1,77 @@
+import React, { useEffect, useState } from "react";
+import { navigate } from "gatsby";
+import { Grid, Link, Menu, NavDropDownButton } from "@trussworks/react-uswds";
+import globe from "../images/logos/globe.png";
+
+function LanguageSelector() {
+  const [isOpen, setIsOpen] = useState(false);
+
+  //Hard coding here for now - will update soon
+  const testMenuItemsOne = [
```
saifulislamdev: The name can be more meaningful.

src/components/language-selector.js
```diff
@@ -0,0 +1,77 @@
...
+            <img src={globe} />
+            <p>Language</p>
+          </div>
+        }
+      />
+      <Menu
+        key="one"
+        items={testMenuItemsOne}
+        isOpen={isOpen}
+        id="languageSelector"
```
saifulislamdev: `language-selector` for naming convention. Change in line 58 as well.

src/components/banner.js
aannyyaa: I'll make two class names - good catch there! I'd just wanted to apply flex to both so reused the name

src/components/section.js
```diff
@@ -1,19 +1,23 @@
...
-        )}
+    <GridContainer className="usa-section">
+      <Grid row>
+        <Grid>
+          {section?.heading && (
+            <h2 className="font-heading-xl">{section.heading}</h2>
+          )}
+          {section?.text && <p className="usa-intro">{section.text}</p>}
+          {buttons?.callToAction && (
+            <Button className="usa-button usa-button--big">
```
saifulislamdev: Is it necessary to assign class `usa-button` if the component is taken from `react-uswds`?

src/components/language-selector.js
aannyyaa: Thanks for pointing out - I actually already updated in my next PR. Just wanted to get the functionality last week!

src/styles/class-overrides.scss
```diff
@@ -0,0 +1,128 @@
+.usa-prose,
```
saifulislamdev: Is there a specific way to organize this? Like in alphabetical order?

src/components/language-selector.js
aannyyaa: Thanks for pointing that out!

src/styles/class-overrides.scss
```diff
@@ -0,0 +1,128 @@
...
+  background-color: color("base-darker");
+}
+
+.usa-nav__primary button[aria-expanded="false"] {
+  color: color("disabled-dark");
+  &:hover {
+    color: color("base-darker");
+  }
+  &::after {
+    border-style: solid;
```
saifulislamny, line 24: Same here, is it possible to organize this? For instance, in alphabetical order? Is there a convention you can follow?

src/styles/class-overrides.scss
```diff
@@ -0,0 +1,128 @@
...
+  }
+}
+
+.usa-button--accent-warm {
+  color: color("base-darker");
+  background-color: color("secondary");
+  &:hover {
+    color: color("base-darker");
+    background-color: color("secondary");
+    border-style: solid;
```
saifulislamdev: I noticed a convention in how you order it `color background-color border-style border-width border-color`, where is this from?

src/components/language-selector.js
aannyyaa: I've been thinking about it and I think we might want the option to have more than a few languages, since it will require a whole other component - the dropdown menu. If the user only wants 2 or 3 languages, then they can just have that number of links next to each other in the banner. But if the user wants to add 13 languages, it would be hard for them to do so since the header wouldn't fit that many links.

aannyyaa: Maybe we should have 2 language selector options for the few and many languages use cases

src/components/section.js
aannyyaa: Yeah because I did custom overrides on the usa-button class

src/styles/class-overrides.scss
aannyyaa: Good point - I'll look into that

src/styles/class-overrides.scss
aannyyaa: I'll look into conventions. I just did text/color related first then border related but didn't consult any documentation regarding ordering

src/components/language-selector.js
saifulislamdev: Rapi mentioned focusing on only English and Spanish. Maybe you can reiterate your idea to him?

src/components/section.js
saifulislamdev: Yes, but I believe `Button` is replaced with `<button class="usa-button"></button>` when it is rendered. Try removing the class name and see if your overrides are still there.

aannyyaa: Yeah seems like the class name isn't needed! I just removed it

src/components/language-selector.js
aannyyaa: Yes let's discuss during our code review tomorrow

saifulislamdev (approved):

src/components/language-selector.js
aannyyaa: Justin would like it if we make it so it automatically renders the language dropdown or language links directly in the header based on how many languages there are. So for now I think it makes sense to work with all of the ones included in the CTO house styles figma

PR #16, opened 2021-07-09
Routing

saifulislamdev:
TODO:
~~* Language switcher does not update if language is entered through the address bar~~


src/components/i18n.js
```diff
@@ -1,16 +1,15 @@
 import i18n from "i18next";
 import Backend from "i18next-xhr-backend";
-import LanguageDetector from "i18next-browser-languagedetector";
+// import LanguageDetector from "i18next-browser-languagedetector";
 import { initReactI18next } from "react-i18next";
 
 i18n
   .use(Backend)
-  .use(LanguageDetector)
+  // .use(LanguageDetector)
```
aannyyaa, line 8: This week let's try to figure out how to use LanguageDetector without it causing routing issues!

src/components/language-switcher.js
```diff
@@ -6,29 +6,26 @@ import "../styles/LanguageSwitcher.css";
...
   const languages = [
     { code: "en", label: "English" },
     { code: "es", label: "Spanish" },
   ];
 
   const handleChange = ({ target }) => {
-    i18n.changeLanguage(target.value);
+    const newLang = target.value;
+    i18n.changeLanguage(newLang);
+    navigate(`/${newLang}/${slug}`);
```
aannyyaa: If you move this to the handle change, does it still navigate to a new page if the language is changed not using the dropdown menu?

src/markdown-pages/landing-es.md
```diff
@@ -1,7 +1,7 @@
 ---
 lang: es
 templateKey: landing-page
-slug: landing
+slug: home
```
aannyyaa, line 4: Is there a specific reason for changing the slug?

src/pages/index.js
```diff
@@ -1,30 +1,57 @@
...
+import { useTranslation } from "react-i18next";
+// import { i18next } from "../components";
 import i18next from "../components/i18n";
+// import NotFound from "../components/404";
 
 import "@trussworks/react-uswds/lib/uswds.css";
 import "@trussworks/react-uswds/lib/index.css";
 
-function Home() {
+function Home({ location }) {
```
aannyyaa, line 8: I'm not sure if we should be redirecting directly from index.js. I think we would probably want to do this in createNodeField in the gatsby-node.js file. Also, in our discussions about routing during our short-term projects, we talked about how using split and selecting a specific hard-coded array element isn't necessarily the way to go with routes. If we change the route structure, then the splitting/selecting/joining logic would all need to be changed too. I'm not super familiar with lodash but maybe we can use that syntax here?

Also, I think we can probably delete index.js altogether. Like I mentioned in code review last week, if slug is 'home' we can go to `/${lang}/` and if slug is not home then we can go to `/${lang}/${slug}/` (this would be in createNodeField in gatsby-node.js).  What do you think?

src/templates/landing-page.js
```diff
@@ -1,33 +1,66 @@
...
+  useEffect(() => {
+    const path = location.pathname;
+    const lang = path.split("/")[1];
+    const route = path
+      .split("/")
+      .slice(2)
+      .filter((v) => v != "")
+      .join("/");
+
+    if (lang.length === 0 && route.length === 0) {
```
aannyyaa, line 28: Thanks for finding a way to change language on the url change! I wonder though if we can utilize language  detector or another i18next feature?

src/pages/404.js
```diff
@@ -0,0 +1,21 @@
+import React, { useEffect } from "react";
+import { Redirect, Router } from "@reach/router";
+import { navigate } from "gatsby";
+import { useTranslation } from "react-i18next";
+import i18next from "../components/i18n";
+
+function NotFound() {
```
aannyyaa, line 7: Thanks for creating a 404 page! Good call :)

src/components/language-switcher.js
aannyyaa: ^ Now I see that you use useEffect in other places. But see my comments on those useEffect hooks below

aannyyaa (changes requested):
Thanks for working on the routing logic! I think we might want to move some of it to gatsby-node.js

src/components/i18n.js
saifulislamdev: Sure!

src/markdown-pages/landing-es.md
saifulislamdev: No major reason, but it looks cleaner.

src/pages/index.js
saifulislamny: From the way you create the markdown pages, we cannot go to `/${lang}/` if the slug is home because the page is loaded on the route `/${lang}/home`.

aannyyaa: Can we change the way it is loaded in the gatsby-node file to load on `/${lang}/` when `slug = home` and load `/${lang}/${slug}/` for any other slug

saifulislamny: Also, handled redirects in `createPages` in the gatsby-node.js file for routes `/en/` and `/es/`.

saifulislamdev: About your `lodash` comment, I'm not sure if we can utilize that or `@reach/router` since they require the location to be matched with the components, but we cannot do that as the component is being used as a template.

saifulislamdev: Changed routing in `gatsby-node.js` with how you said it. Amazing! If the slug is 'home', we go to `/${lang}/` and if the slug is not 'home', then we go to `/${lang}/${slug}/`. At first, I had read your comment wrong.

src/templates/landing-page.js
saifulislamdev: Hmm, from what I have seen in the documentation before, I don't think so. Does anything stick out to you?

src/pages/index.js
saifulislamdev: Handled redirects in `createPages` in the gatsby-node.js file for routes `/en/home` and `/es/home` so the only redirecting in `index.js` is for the `/` route which gets redirected to `/en` or `/es` based on the language.

aannyyaa: Awesome!!! And okay good to know about the issues with location + templates

src/templates/landing-page.js
aannyyaa: I'll try to look into the docs more, but when I looked a week or two ago I didn't find any good examples

aannyyaa (approved):

PR #19, opened 2021-07-14
NYC Footer (Return to top)

saifulislamdev:


src/components/footer.js
```diff
@@ -13,25 +14,18 @@ import logoImg from "../images/logos/cto_logo_horizontal_white.png";
 function Footer() {
   const { t } = useTranslation();
   const primaryLinks = ["/link-one", "/link-two", "/link-three", "/link-four"];
```
aannyyaa, line 16: I think the links should be in the translation.json or a file in the constants folder. Probably the translation.json since other information about the links is there

src/components/footer.js
```diff
@@ -13,25 +14,18 @@ import logoImg from "../images/logos/cto_logo_horizontal_white.png";
...
   return (
     <FooterUSWDS
       size="medium"
-      // returnToTop={returnToTop}
+      returnToTop={
+        <GridContainer>
+          <Link to="#top" className="usa-footer__return-to-top">
+            Return to top
+          </Link>
+          {/* TODO: for some reason class name has to be manually added*/}
```
aannyyaa, line 26: What do you mean by this? That you need to add `className="usa-footer__return-to-top"` to see the proper styling?

src/components/footer.js
```diff
@@ -13,25 +14,18 @@ import logoImg from "../images/logos/cto_logo_horizontal_white.png";
...
-    "rss",
-  ];
 
   return (
     <FooterUSWDS
       size="medium"
-      // returnToTop={returnToTop}
+      returnToTop={
+        <GridContainer>
+          <Link to="#top" className="usa-footer__return-to-top">
```
aannyyaa, line 23: Nice!

src/styles/class-overrides.scss
```diff
@@ -97,6 +97,12 @@
   }
 }
 
+.usa-footer__return-to-top {
+  color: color("base-darker");
+  line-height: $mocto-box-height;
```
aannyyaa: I'm wondering why you set line-height here? Also in my most recent PR I got rid of the following in the mocto-theme file: 
```
$mocto-button-border-width: 2px;
$mocto-link-border-width: 4px;
$mocto-link-padding: 2px;
```
instead I use spacing tokens in custom overrides, like this: `padding-left: units("105");` I think it makes sense to use a spacing token here

src/styles/mocto-theme.scss
```diff
@@ -21,6 +21,7 @@ $mocto-disabled-light: #E9E9E9;
 $mocto-disabled: #C4C4C4;
 $mocto-disabled-dark: #7B7B7B;
 
+$mocto-box-height: 19px;
```
aannyyaa: Building on my comment above, I think we should use spacing tokens. Also is there a specific reason why you used `19px`? We should probably stick to the units used in USWDS. We could use the token `"205"` which corresponds to `20px`.

aannyyaa (changes requested):

src/styles/mocto-theme.scss
aannyyaa: Also [here](https://designsystem.digital.gov/design-tokens/spacing-units/) is the link to the spacing units guide from USWDS. I know it's sometimes difficult to track down the right page in the docs

src/components/footer.js
saifulislamdev: Keeping it in `translation.json` would be a duplication because the user will be directed to the same route no matter what language they're reading. But file in the constants folder, good idea!

src/components/footer.js
saifulislamdev: Yes, it's weird because there's a `returnToTop` attribute that automatically includes the class, but it doesn't here. It works in LPaaS 1.0, but I cannot see how they value that attribute. I think it's its own component, but it still does not include the class when I have it as a separate component.

aannyyaa: Oh interesting!

src/styles/class-overrides.scss
saifulislamdev: Got rid of `line-height`. I used spacing tokens in custom overrides otherwise.

src/styles/mocto-theme.scss
saifulislamdev: Spacing tokens are used now, didn't need to use `$mocto-box-height` for styling regardless. I knew about the tokens but didn't think we'd need them, thanks!

aannyyaa (approved):

PR #21, opened 2021-07-14
Styling Updates

aannyyaa:


src/components/language-selector.js
```diff
@@ -16,6 +16,7 @@ function LanguageSelector({ slug }) {
   };
 
   const languageButtonGroup = languages.map((language) => (
+    <div className="banner__language-selector-item">
```
saifulislamny, line 19: Do you need to create a new `div`, or can you add a `class` attribute in the inner `Button`?

saifulislamdev: Do you need to create a new `div`, or can you add a `class` attribute in the inner `Button`?

aannyyaa: I used div here and for the language buttons in the dropdown menu because the border that appears on hover has slanted edges for the buttons, so this was my work around. If you can figure out how to style just one side of the border and not have it slanted, then we can definitely remove the div

saifulislamdev: The only possible suggestion I have is to have all the buttons wrapped by only one div with the class. I don't know if it will work or is against a best practice. Let me know what you think; I'm down to leave it like this.

saifulislamdev (approved):

PR #22, opened 2021-07-14
NYC Links

saifulislamdev:


src/styles/mocto-theme.scss
```diff
@@ -21,9 +21,12 @@ $mocto-disabled-light: #E9E9E9;
 $mocto-disabled: #C4C4C4;
 $mocto-disabled-dark: #7B7B7B;
 
+$mocto-box-height: 19px;
 $mocto-button-border-width: 2px;
-$mocto-link-border-width: 4px;
-$mocto-link-padding: 2px;
+$mocto-link-border-width: 4px; /* TODO: renaming needed so this naming can be used for actual links */
+$mocto-link-padding: 2px; /* TODO: renaming needed so this naming can be used for actual links */
```
saifulislamdev: You use these tokens for the menu in the mobile view, maybe consider renaming them so it's not confused for the actual links?

src/styles/mocto-theme.scss
```diff
@@ -21,9 +21,12 @@ $mocto-disabled-light: #E9E9E9;
 $mocto-disabled: #C4C4C4;
 $mocto-disabled-dark: #7B7B7B;
 
+$mocto-box-height: 19px;
```
saifulislamdev, line 24: This is also in another PR.

src/styles/mocto-theme.scss
aannyyaa: I haven't specifically worked on the mobile view yet. 

Also, I mentioned this in another PR, but I no longer use mocto variables for spacing in the theme file. I have been able to just use spacing tokens in the class overrides file. I think doing that here would make sense

src/styles/mocto-theme.scss
aannyyaa: I made a comment in your other PR about this - it's also related to the comment above about using spacing tokens

src/styles/class-overrides.scss
```diff
@@ -35,6 +35,36 @@
   }
 }
 
+.usa-link {
```
aannyyaa, line 107: Looking good! We can update this to use spacing tokens instead of `$mocto-link-width`, etc. I also started using mixins to save us from repeating code. It might be useful to use a mixin here since `.usa-link` and `.usa-link--light` are similar

aannyyaa (changes requested):

src/styles/mocto-theme.scss
saifulislamdev: Spacing tokens in the class overrides file completed! MOCTO variables are not used for spacing anymore. The other variables will be dealt with in #19.

src/styles/mocto-theme.scss
saifulislamdev: I will adjust this in the other PR.

src/styles/class-overrides.scss
saifulislamdev: Spacing tokens and mixins used!

aannyyaa (approved):

PR #24, opened 2021-07-16
Refactoring translation files and dynamically create header items

saifulislamdev:
Delete the older PR when you'd like to, kept it there if you wanted to refer to it.

### DIFF
--- a/src/components/header.js
+++ b/src/components/header.js
@@ -45,7 +45,7 @@ function Header({ slug }) {
       ? constFileLength
       : translationFileLength; // take shorter length in case there is missing dropdowns in `../constants/links.js` (constants file) or `../locales/` (translation files)
   if (translationFileLength !== constFileLength)
-    console.warn(
+    console.error(
       "Different number of dropdowns in /src/constants/link.js and dropdown labels in /src/locales\n",
       `${constFileLength} dropdown${
         constFileLength > 1 ? "s" : ""
@@ -65,7 +65,7 @@ function Header({ slug }) {
           ? navDropdownLinkLabelsLength
           : navDropdownLinksLength; // take shorter length in case there is a missing link in `../constants/links.js` (constants file) or label in  `../locales/` (translation files)
       if (navDropdownLinksLength !== navDropdownLinkLabelsLength)
-        console.warn(
+        console.error(
           `Different number of links in /src/constants/link.js and link labels in /src/locales for dropdown ${
             i + 1
           }\n`,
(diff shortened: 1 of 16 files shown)


src/components/header.js
```diff
@@ -32,29 +34,90 @@ function Header({ slug }) {
...
+  let dropdowns = [];
+  const dropdownLinks = links.navDropdowns;
+  const constFileLength = dropdownLinks.length;
+  const dropdownLabels = t("header.nav.dropdowns");
+  const translationFileLength = dropdownLabels.length;
+  const length =
+    translationFileLength > constFileLength
+      ? constFileLength
+      : translationFileLength; // take shorter length in case there is missing dropdowns in `../constants/links.js` (constants file) or `../locales/` (translation files)
+  if (translationFileLength !== constFileLength)
```
aannyyaa, line 47: Nice adding the warning! Although, I was wondering if it should be an error instead of a warning?

src/components/header.js
```diff
@@ -32,29 +34,90 @@ function Header({ slug }) {
...
+  const parentLinkItems = parentLinks.map((element, i) => {
+    return (
+      i < parentLength && (
+        <Link to={element} variant="nav" key={i}>
+          <span>{parentLinksLabels[i]}</span>
+        </Link>
+      )
+    );
+  });
   const navBarItems = [
```
aannyyaa, line 121: `navBarItems` can be created dynamically too by looping through `dropdowns`. This would be useful if the user wanted more or less than 2 nav dropdowns

src/constants/links.js
```diff
@@ -0,0 +1,11 @@
+export const header = {
+  navDropdowns: [
+    ["link-one", "link-two"],
```
aannyyaa, line 3: Should these paths have `/`?

src/components/header.js
saifulislamdev: Good idea. After all, they are going to get unexpected behavior. I didn't want to give an error message at first because I thought it would crash the whole site, but after adding `console.error`, I see it still loads the whole page, but with a different type of message in the console.

src/components/header.js
saifulislamdev: This will be handled in another PR.

src/constants/links.js
saifulislamdev: Good catch! But it should be how it is right now, otherwise, it would redirect to `/link-one` instead of `/en/link-one`. The footer links don't have any `/` now.

PR #25, opened 2021-07-16
Airbnb ESLint config

aannyyaa:


src/components/header.jsx
```diff
@@ -1,28 +1,21 @@
-import React, { useState } from "react";
```
saifulislamdev: How will my changes in `header.js` and `footer.js` be transferred over? For `header.js`, it looks like it may be transferred over, but for `footer.js` it seems like you're deleting the file entirely and creating a new one.

src/templates/landing-page.jsx
```diff
@@ -36,9 +44,9 @@ function Landing({ data, location }) {
       // only route given
       // not currently being used
       // navigate(`${i18n.language}/${route}`);
-    } else {
+    } else if (lang !== i18n.language) {
```
saifulislamdev, line 47: Nice catch!

src/components/header.jsx
aannyyaa: I renamed all the files that use JSX by using the syntax `git mv oldfile.js newfile.jsx`. Following the airbnb eslint guidelines in many cases two quotation marks are changed to one. Maybe if there were changes throughout the whole file and the file was moved it says it was deleted? Not sure why it considers footer deleted but not header

src/templates/landing-page.jsx
aannyyaa: I wish I caught it myself but it was airbnb eslint!

src/components/header.jsx
saifulislamdev: Let's merge this after my PRs are merged.

src/components/footer.jsx
```diff
@@ -0,0 +1,67 @@
...
+  Logo,
+} from '@trussworks/react-uswds';
+import { useTranslation } from 'react-i18next';
+import { Link } from '.';
+
+import logoImg from '../images/logos/cto_logo_horizontal_white.png';
+
+function Footer() {
+  const { t } = useTranslation();
+  const primaryLinks = ['/link-one', '/link-two', '/link-three', '/link-four'];
```
saifulislamdev, line 16: I think it would be best to leave out the changes in the components until my PRs are merged. Once they're merged, you can merge dev into your branch to make sure the changes are made in the components and then create the necessary .jsx files.

saifulislamdev (approved):

PR #28, opened 2021-07-19
Update Link Styling

aannyyaa:


src/styles/class-overrides.scss
```diff
@@ -44,11 +44,13 @@
...
 @mixin link-size {
-  border-bottom-width: units('1px') * 3;
+  // border-bottom-width: units('1px') * 3;
+  border-bottom-width: units('05');
 }
 
 @mixin link-size-expanded {
-  border-bottom-width: units('1px') * 10;
+  // border-bottom-width: units('1px') * 10;
+  border-bottom-width: units('105');
```
saifulislamdev, line 53: Thanks for solving the conflicts! I took this directly from the Figma, and there were no conflicts in the branch I was working in, so I didn't have to address these.

src/styles/class-overrides.scss
```diff
@@ -116,11 +122,15 @@
   border-bottom-style: solid;
   border-bottom-color: color('primary');
   &:hover {
+    text-decoration: none;
```
saifulislamdev, line 125: Is this necessary if it's already in line 120?

saifulislamdev (approved):

saifulislamdev (changes requested):
Only one minor change, if needed.

src/styles/class-overrides.scss
aannyyaa: These weren't conflicts - I was just trying to use the USWDS units. So they are slightly different sizes than in the figma

src/styles/class-overrides.scss
aannyyaa: Without this line, the white underline still showed up on hover in addition to the bottom border

src/styles/class-overrides.scss
saifulislamdev: I was referring to the overlap you mentioned earlier.

saifulislamdev (approved):

PR #29, opened 2021-07-19
NavDropDown

saifulislamdev:


src/components/nav-drop-down.jsx
```diff
@@ -0,0 +1,99 @@
+import React, { useState } from 'react';
```
aannyyaa, line 1: Good idea to create a new component for the nav dropdown! I was just thinking about that too

src/components/nav-drop-down.jsx
```diff
@@ -0,0 +1,99 @@
...
+  const navBarItems = dropdowns
+    .map((_, i) => {
+      const id = `navDropDown${i}`;
+      const iStr = i.toString();
+      /* dynamically create dropdowns in navigation bar */
+      return (
+        <>
+          <NavDropDownButton
+            onToggle={() => {
+              if (open === id) setOpen(undefined);
```
aannyyaa, line 80: Why do you set open to `undefined` instead of `false`?

src/components/nav-drop-down.jsx
```diff
@@ -0,0 +1,99 @@
...
+            {navDropdownLinkLabels[j]}
+          </Link>
+          ),
+        ),
+      );
+    } else dropdowns.push([]);
+  });
+
+  /* dynamically create dropdowns */
+  const [open, setOpen] = useState();
```
aannyyaa: Kudos on dynamically creating the dropdowns!!

src/components/nav-drop-down.jsx
saifulislamdev: It works both ways, but I used `undefined` because what's stored in the state is the `id` of the dropdown currently open. An`id` corresponds to being `undefined` more than it does to be `false`, so I went with `undefined`.  Is it preferred to use false?

src/components/nav-drop-down.jsx
saifulislamdev: It was a process with one thing after another, but glad it's finally solved.

src/components/nav-drop-down.jsx
aannyyaa: Oh okay, you're setting the id to undefined. I had been thinking of it as true/false for whether it is open or not.

Also when you use `useState` should you set a default state? (on line 70)

saifulislamdev: Default state set!

PR #30, opened 2021-07-19
Mobile Responsive Header & Language Selector

aannyyaa:


aannyyaa:
TODO: 

- styling languages in nav drawer
- styling other items in nav drawer

src/components/header.jsx
```diff
@@ -49,6 +52,94 @@ function Header({ slug }) {
     ),
   );
 
+  const handleClick = (langKey) => {
+    i18n.changeLanguage(langKey);
+    navigate(`/${langKey}/${slug}`);
+  };
+
+  const languageNavItems = (
```
aannyyaa, line 77: Will probably want to create a new component for this

src/styles/_class-overrides.scss
```diff
@@ -315,3 +321,137 @@
 .graphic_image--circle {
   border-radius: 50%;
 }
+
+//mobile & tablet view
+@media (max-width: 1023px) {
```
aannyyaa: In a future PR I'll try to clean up the media queries using more custom mixins or functions

aannyyaa: I used to have `min-width: units('mobile')`, but then realized that there were multiple things I'd have to reset in desktop view so it seemed to make more sense for now to use `max-width`. However, `max-width: units('mobile')` didn't seem like the best idea since at 1024px some of the desktop specific styles started but mobile styles would also be applied if they weren't over-ridden by the desktop styles. I tried  `max-width: units('mobile') - 1px`, but got an error saying units('mobile') is in rem not px

src/components/header.jsx
```diff
@@ -49,6 +52,52 @@ function Header({ slug }) {
...
+    <div className="header__language-nav-container">
+      <div className="header__language-nav-RtoL">
+        {languages
+          .filter((language) => language.isRtoL)
```
saifulislamdev: Do you have to filter then map? I think you can use a conditional in the map instead. It might look more readable the way you have it now, but you're doing more operations currently.

src/components/header.jsx
```diff
@@ -49,6 +52,52 @@ function Header({ slug }) {
...
+                key={language.langKey}
+              >
+                {language.lang}
+              </Button>
+            </div>
+          ))}
+      </div>
+      <div className="header__language-nav-LtoR">
+        {languages
+          .filter((language) => !language.isRtoL)
```
saifulislamdev: Same there here.

src/components/header.jsx
```diff
@@ -49,6 +52,52 @@ function Header({ slug }) {
...
+          .map((language) => (
+            <div className="header__language-nav-items">
+              <Button
```
saifulislamdev: Seems like you're repeating the code over here, try not to.

src/styles/_class-overrides.scss
```diff
@@ -0,0 +1,525 @@
...
+            padding: units(2);
+            margin: 0;
+            border: none;
+          }
+        }
+      }
+    }
+  }
+
+  /*hide language buttons from nav in desktop view
```
saifulislamdev, line 402: The `mocto-theme.scss` file was renamed, but this shows up as a completely new file, so I can't identify the changes. What changes did you make?

aannyyaa: Mostly just the media queries - line 325 on. I'm planning to make some functions/mixins to make the code more concise and readable

src/styles/_class-overrides.scss
saifulislamdev: I got the same error too when trying to do something similar. It seems to happen when performing an operation with two or more non-scalar values. You might remember me doing something like `units('1px') * 3` which solved the problem. Try to look into how you can store the values in variables and then perform the operation, or convert the units('mobile') to pixels somehow using an SCSS function we might not know about.

src/styles/_class-overrides.scss
```diff
@@ -0,0 +1,525 @@
...
+  //languages in nav bar
+  .header__language-nav-container {
+    //position languages at the bottom
+    position: absolute;
+    bottom: 0;
+    @include u-bg('base-dark');
+    padding: units(1) 0;
+    //right align right to left languages
+    .header__language-nav-RtoL {
+      display: flex;
```
saifulislamdev: I know you were going to use mixins, but one can be done here.

src/styles/_class-overrides.scss
```diff
@@ -0,0 +1,525 @@
...
+    padding: units(2);
+    .usa-nav__inner {
+      .usa-nav__primary-item {
+        .usa-nav__submenu {
+          @include u-bg('base-darker');
+          .usa-nav__submenu-item {
+            @include u-bg('base-darker');
+            a {
+              @include u-text('accent-cool-lighter');
+              @include u-bg('base-darker');
```
saifulislamdev, line 351: `@include u-bg('base-darker')` is repeated multiple times in lines 455, 457, and 460. Does it have to be?

src/styles/_class-overrides.scss
```diff
@@ -0,0 +1,525 @@
...
+            @include u-text('bold');
+            padding: units(2);
+            margin: 0;
+            border: none;
+          }
+        }
+        .usa-accordion__button[aria-expanded='false'] {
+          @include u-text('disabled-dark');
+          @include u-bg('accent-cool-lighter');
+          @include u-text('bold');
```
saifulislamdev: Mixins can be used here.

src/styles/_class-overrides.scss
aannyyaa: It's okay. This is outdated now - I switched back to `min-width: units('mobile')` yesterday in the commit `use min-width for mobile view, update desktop view`

src/styles/_class-overrides.scss
aannyyaa: As things are now, if I only set the background color on 455 then the backgrounds in the other nested elements are white, since they were set to that color in the mobile view. I think each has to be reset for desktop view?

src/components/header.jsx
aannyyaa: Changed to conditional in map

src/styles/_class-overrides.scss
saifulislamdev: I see, is there an element like "others" so you don't have to repeat it?

src/components/header.jsx
```diff
@@ -49,20 +52,72 @@ function Header({ slug }) {
...
+            </Button>
+          </div>
+          ),
+        )}
+      </div>
+      <div className="header__language-nav-LtoR">
+        {languages.map(
+          (language) => !language.isRtoL && (
+          <div className="header__language-nav-items">
+            <Button
```
saifulislamdev: Repetition still going on from lines 84 -93.

aannyyaa: Updated to use a function to return language nav items and avoid repetition

src/styles/_class-overrides.scss
aannyyaa: I tried to use `@extend`, but got this error:
```
You may not @extend an outer selector from within @media.
```

src/styles/_class-overrides.scss
aannyyaa: done

src/styles/_class-overrides.scss
aannyyaa: done

saifulislamdev: There are repetitions in lines 403-407 and 411-415, should we leave that as it is?

src/styles/_class-overrides.scss
saifulislamdev: There are repetitions in lines 467-471, 473-477, 482-486, 488-492, 500-504, and 506-510. Should we leave those?

aannyyaa: I think we can definitely work on cleaning up repetition once we have all the features and styling we want

saifulislamdev (approved):

PR #31, opened 2021-07-20
Coding Agreements

saifulislamdev:
1. Organization of file imports through ESLint plugin and configuration
1. File naming for images: separated by underscores (already set)
1. `.map()` > `.from()` used throughout entire app
1. Single quotes instead of double quotes (unless for HTML or required)
1. General order for HTML attributes: event handlers, type, id/for, name, class, href/src, styling, key 
1. General order of class names: Alphabetical order ~~Order of least importance to most importance~~
1. `function` instead of `const` for components
1. `isOpen` not `open` in `src/components/nav-drop-down.jsx`
1. General order of props: alphabetical if a short list of props and importance if many props
1. …DropDowns… > …Dropdowns… (i.e. navDropDowns > navDropdowns)
1. General order in translation files: in order of view (things on the top of the page come first, the bottom of the page come last)
1. General order in `src/components/i18n.js`: alphabetical order of what’s being used and then alphabetical order of options in init
1. General order in `.gitignore`: order of importance
1. General order in `gatsby-node`: alphabetical order
1. General order in SCSS styling: mixins, classes, media queries (subject to change)


src/components/footer.jsx
```diff
@@ -35,6 +35,7 @@ function Footer() {
       '\n',
     );
   }
+  // TODO: link underline is longer now after making it .map
```
aannyyaa, line 38: I actually just added link underline length as a mini-task in styling clean up. The lengths were also too long in mobile view in the footer

aannyyaa (approved):

.gitignore
```diff
@@ -1,4 +1,9 @@
-node_modules/
-.DS_Store
+# Gatsby
 .cache/
-public/
\ No newline at end of file
+public/
+
+# React
+node_modules/
```
aannyyaa: why are node modules under react?

src/components/banner.jsx
```diff
@@ -1,8 +1,10 @@
 import React from 'react';
-import PropTypes from 'prop-types';
+
 import { Grid, GridContainer } from '@trussworks/react-uswds';
-import { LanguageSelector } from '.';
+import PropTypes from 'prop-types';
```
aannyyaa, line 4: What is the ordering of imports?

src/components/graphic.jsx
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
-import PropTypes from 'prop-types';
+
 import { Grid, GridContainer } from '@trussworks/react-uswds';
+import PropTypes from 'prop-types';
 
 function Graphic({ graphics }) {
```
aannyyaa, line 6: What is the reasoning behind "function instead of const for components" as you said in your comment on the PR?

.gitignore
saifulislamdev: Good point. Can you think of a better way to name it?

src/components/banner.jsx
saifulislamdev: It's according to the configuration set in `.eslintrc`. 
* Built-in imports, then external and internal.
* The React import comes before everything.
* There are new lines between each group.
* In each group, imports are alphabetized.

src/components/graphic.jsx
saifulislamdev: To distinguish when new components are being made. You may have another function in the same file, but it can be named `const` to distinguish the actual component from a helper function.

.gitignore
aannyyaa: either it could just go at the top of the file and have no name or it could be under `General`?

src/components/banner.jsx
aannyyaa: Okay awesome

src/components/graphic.jsx
aannyyaa: What resources suggest doing that?

.gitignore
saifulislamdev: Done.

src/components/graphic.jsx
saifulislamdev: No resources were used; it's something I thought of myself.

aannyyaa: Okay - I think it would be better to consistently use es6 syntax

PR #32, opened 2021-07-20
Translation Refactoring

aannyyaa:


src/components/footer.jsx
```diff
@@ -47,7 +47,7 @@ function Footer() {
               <img
                 className="footer__logo"
                 src={logoImg}
-                alt={t('footer.secondary.heading.logo')}
+                alt={t('agency.longformName')}
```
saifulislamny, line 50: What does longform mean?

saifulislamdev, line 52: Minor adjustment: `agency.longFormName`

I didn't know what longform meant at first.

src/components/header.jsx
```diff
@@ -105,10 +105,14 @@ function Header({ slug }) {
       <Banner slug={slug}>{t('header.banner')}</Banner>
       <div className="usa-navbar">
         <div className="header__logo-title">
-          <img className="header__logo" src={ctoLogoShortform} alt="NYC CTO" />
+          <img
+            className="header__logo"
+            src={ctoLogoShortform}
```
saifulislamdev: Minor adjustment: `ctoLogoShortForm`

src/components/header.jsx
```diff
@@ -105,10 +105,14 @@ function Header({ slug }) {
...
+            alt={t('agency.shortformName')}
```
saifulislamdev, line 111: Minor adjustment: `agency.shortFormName`

src/locales/ar/translation.json
```diff
@@ -1,5 +1,10 @@
 {
   "title": "[ar] Project Title",
+  "agency": {
```
saifulislamdev, line 3: Why is this outside of the footer brace? Will it be used throughout the page?

src/components/banner.jsx
```diff
@@ -1,18 +1,20 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { useTranslation } from 'react-i18next';
 import { Grid, GridContainer } from '@trussworks/react-uswds';
 import { LanguageSelector } from '.';
 import nycLogo from '../images/logos/nyc_logo.png';
```
saifulislamdev: Should more general file names be used for images (i.e.`logo.png`)? It will make our app more generalized (not only for CTO and NYC agencies).

src/components/header.jsx
aannyyaa: I was trying to be consistent with the names of the logos from the mocto website which are like `cto_logo_shortform_dark` and are treating shortform and longform as 1 word

src/locales/ar/translation.json
aannyyaa: Yeah I used it in header too for the alt tags - it seems like something that could be used in multiple places

src/locales/ar/translation.json
```diff
@@ -1,5 +1,10 @@
...
+    "longformName": "[ar] NYC Mayor's Office of the Chief Technology Officer",
+    "shortformName": "[ar] NYC CTO"
+  },
+  "nyc": "[ar] NYC",
```
saifulislamdev: A more general property could be used instead of `nyc`, maybe `location`? Also, why is this outside of the `footer` brace?

src/components/banner.jsx
aannyyaa: The image is the nyc logo, so I don't think it makes sense to change the file name. We could change the variable name from `nycLogo` to `logo` though

src/locales/ar/translation.json
aannyyaa: It's used in the banner for an alt tag, not in the footer. A more general key is a good idea

src/locales/es/translation.json
```diff
@@ -1,5 +1,10 @@
 {
   "title": "[es] Project Title",
+  "department": {
```
saifulislamdev: You name it `department` instead of `agency` here.

aannyyaa: good catch

src/components/header.jsx
saifulislamdev: Should we change the file and variable names so that it is more generalized for agencies?

src/components/banner.jsx
saifulislamdev: What if this app becomes more readily used by the public (not only City agencies)? Then, they would have to make their file name `nyc_logo` as well for this to work. Otherwise, they would have to go into the code and change it themselves manually.

src/components/header.jsx
aannyyaa: I'm updating the variable names now. I was thinking maybe we can have a file like `src/index.js` that exports all of the images, and that would be the one place where the image paths need to change based on the file names

saifulislamdev: Great idea, go for it!

src/components/banner.jsx
aannyyaa: See my comment on header.jsx about creating a file for image exports. Even if we change this file to logo, then we're still asking the user to update their file names to match ours exactly. We can have the exports file with general variable names and where the user just updates the paths to match their file names?

aannyyaa: I changed to more general file names, as you suggested, and also included the `images/index.js` file for exporting images

saifulislamdev (approved):

PR #33, opened 2021-07-22
Mobile Responsive Footer

aannyyaa:


saifulislamdev (approved):

saifulislamdev:
Try to think more about how you can potentially organize your `_class-overrides` code.

saifulislamdev:
We might have discussed this already, but how come all the class overrides are in `_class-overrides`? Can we break it down into different files? It's currently ~600 lines.

aannyyaa:
> We might have discussed this already, but how come all the class overrides are in `_class-overrides`? Can we break it down into different files? It's currently ~600 lines.

We did discuss in code review and said then it could stay in one file. Maybe it would make sense though to have a mixins file, general styles (links, buttons, etc) file, and media queries file?

saifulislamdev:
> We did discuss in code review and said then it could stay in one file. Maybe it would make sense though to have a mixins file, general styles (links, buttons, etc) file, and media queries file?

I thought it could follow the structure we have for our components so that the styles match each component, and it would be easy to find. It's commonly done this way as well unless USWDS projects follow a certain format.

PR #35, opened 2021-07-23
Documentation Page Template

aannyyaa:


src/styles/index.scss
```diff
@@ -8,9 +8,10 @@ $theme-font-path: '../../node_modules/uswds/src/fonts';
...
 //Custom themes
-@import './_custom-theme.scss';
+@import './custom-theme.scss';
 
 @import 'uswds';
 
 //Class overrides and custom classes
-@import './_class-overrides.scss';
+@import './class-overrides.scss';
+@import './documentation-page'
```
saifulislamdev, line 17: I see you don't use `.scss` here. Do you need to? If not, stick to a convention in this file of including it or not at the end of imports.

src/templates/documentation-page.jsx
```diff
@@ -0,0 +1,89 @@
+/* eslint-disable react/no-danger */
+import React, { useEffect } from 'react';
+import PropTypes from 'prop-types';
+import { Helmet } from 'react-helmet';
+import { graphql } from 'gatsby';
+import { I18nextProvider, useTranslation } from 'react-i18next';
+import { Grid, GridContainer } from '@trussworks/react-uswds';
+import { i18next, Layout } from '../components';
+
+import '@trussworks/react-uswds/lib/uswds.css';
```
saifulislamdev, line 10: Do these imports have to be repeated for each template? If so, you might want to avoid this repetition.

src/templates/documentation-page.jsx
```diff
@@ -0,0 +1,89 @@
...
+    }
+  }, [location.pathname]);
+
+  const { markdownRemark } = data;
+  const { frontmatter, html } = markdownRemark;
+
+  return (
+    <>
+      <I18nextProvider i18n={i18next}>
+        <Helmet title={t('title')} htmlAttributes={{ lang: i18n.language }} />
```
saifulislamdev: It seems like you're taking the title of the page from the translation files. It might be a good time to start having the titles in the markdown files instead.

src/styles/index.scss
aannyyaa: Good point - seems like it's not needed

src/templates/documentation-page.jsx
aannyyaa: Yes otherwise the uswds styles are not applied

saifulislamdev: Can you think of a way to avoid repetition? We can, of course, put in the documentation a proposed template for templates that would include these imports. However, is there a way we can avoid repeating?

aannyyaa: Yeah my only thought was a template/layout component for templates that includes all the imports

saifulislamdev: Sounds good!

src/markdown-pages/documentation-zh_HANS.md
```diff
@@ -0,0 +1,26 @@
+---
+lang: zh_hans
+templateKey: documentation-page
+slug: documentation
+pageTitle: Project | Documentation
```
saifulislamdev: Do you want to add placeholders for the page title (i.e. `[zh_hans] Project | Documentation`) across the different files?

src/templates/documentation-page.jsx
```diff
@@ -0,0 +1,95 @@
...
+    <>
+      <I18nextProvider i18n={i18next}>
+        <Helmet
+          title={frontmatter.pageTitle}
+          htmlAttributes={{ lang: i18n.language }}
+        />
+        <Layout slug={frontmatter.slug}>
+          <main>
+            <GridContainer>
+              <Grid className="documentation__container">
```
saifulislamdev, line 58: The naming of this class may be misinterpreted. If you decide to change this, make sure to change the class name in the overrides accordingly.

src/templates/documentation-page.jsx
```diff
@@ -0,0 +1,95 @@
...
+                <div
+                  className="documentation usa-layout-docs__main usa-prose usa-layout-docs"
```
saifulislamdev: Alphabetical order as decided in coding agreements.

src/templates/documentation-page.jsx
aannyyaa: In what way would it be misinterpreted?

src/markdown-pages/documentation-zh_HANS.md
aannyyaa: Yeah I did that for the landing page but forgot to for the documentation page - just added it

src/templates/documentation-page.jsx
saifulislamdev: The class name is `...container`, but it's in a `Grid` element, not the `GridContainer` element. I might be thinking too much into it – resolving this.

src/templates/documentation-page.jsx
aannyyaa: fixed

PR #36, opened 2021-07-23
Feedback Module Placement

aannyyaa:


src/components/layout.jsx
```diff
@@ -18,8 +17,6 @@ function Layout({ children, slug }) {
       <Helmet onChangeClientState={(newState) => setPageTitle(newState.title)} />
```
saifulislamdev: This line should be available in the other file.

src/styles/_class-overrides.scss
```diff
@@ -347,6 +347,10 @@
     padding: 0;
   }
 
+  #feedback-widget{
```
saifulislamdev: I know this is fine, but Justin said we should stick to using classes only. Do you think instead we should add a class name and use it here?

aannyyaa: Yeah I just quickly wanted to update the styling before the show and tell, and there was already an id. I updated it to include a class too

src/components/layout.jsx
aannyyaa: I moved it to footer, but I'm wondering why we need to update the helmet title? We already have helmet in our template pages. Would having it here interfere?

aannyyaa: Can we use the title from our translation file and have it like this:

```
<div id="feedback-widget" className="footer__feedback-widget" lang={i18n.language} pageTitle={t('title')} endpoint={process.env.GATSBY_ENDPOINT} />

```

Specifically, use `pageTitle={t('title')}`

aannyyaa: Also why do we need:  
```
const [pageTitle, setPageTitle] = useState(t('title')); // initial title is header title
```

saifulislamdev: > I moved it to footer, but I'm wondering why we need to update the helmet title? We already have helmet in our template pages. Would having it here interfere?

I use the Helmet to update the state `pageTitle`. Also, don't see why it would interfere with the Helmet in the template pages since I'm not changing anything about the page; I'm only updating the state.

saifulislamdev: > Can we use the title from our translation file and have it like this:
> 
> ```
> <div id="feedback-widget" className="footer__feedback-widget" lang={i18n.language} pageTitle={t('title')} endpoint={process.env.GATSBY_ENDPOINT} />
> ```
> 
> Specifically, use `pageTitle={t('title')}`

I'm not sure if we should use the default title because each page can have a different title and the default title won't be as specific compared to using the page title.

saifulislamdev: > Also why do we need:
> 
> ```
> const [pageTitle, setPageTitle] = useState(t('title')); // initial title is header title
> ```

We need `useState` to update the title for the feedback module for every page change.

aannyyaa: If we set page title in the markdown, we can use it to set the title in helmet and for the feedback module. I don't think we need `useState` in that case.

aannyyaa: Also you use the page title for the initial state, I don't see when that state would change. Won't the feedback module be rendered on every page, which are all created at build time?

aannyyaa: > > I moved it to footer, but I'm wondering why we need to update the helmet title? We already have helmet in our template pages. Would having it here interfere?
> 
> I use the Helmet to update the state `pageTitle`. Also, don't see why it would interfere with the Helmet in the template pages since I'm not changing anything about the page; I'm only updating the state.

But is it okay to use the Helmet component multiple times, in a nested way? Might you be updating the title only on the Helmet in the footer (or used to be in the layout) and not the one in the template page?

saifulislamdev: > If we set page title in the markdown, we can use it to set the title in helmet and for the feedback module. I don't think we need `useState` in that case.

Definitely! But currently, we don't have the page title in the markdown.

saifulislamdev: > Also you use the page title for the initial state, I don't see when that state would change. Won't the feedback module be rendered on every page, which are all created at build time?

That state would change in the `onChangeClientState` attribute of the Helmet.

saifulislamdev: > > > I moved it to footer, but I'm wondering why we need to update the helmet title? We already have helmet in our template pages. Would having it here interfere?
> > 
> > 
> > I use the Helmet to update the state `pageTitle`. Also, don't see why it would interfere with the Helmet in the template pages since I'm not changing anything about the page; I'm only updating the state.
> 
> But is it okay to use the Helmet component multiple times, in a nested way? Might you be updating the title only on the Helmet in the footer (or used to be in the layout) and not the one on the template page?

I see what you're saying, but it depends on when the client state changes to trigger the event handler. For example, if `onChangeClientState` is called on a title change, we wouldn't have to worry about it. But if it does not trigger on a title change, then I think the `Helmet` should be in the template, and the same title passed into the `Helmet` in the template should be passed to the `Layout` -> `Footer` -> `title` in Feedback Module. We can message each other for further clarification if needed.

src/components/footer.jsx
```diff
@@ -55,6 +56,8 @@ function Footer() {
           <Link to="#top">
             {t('footer.returnToTop')}
           </Link>
+          <div id="feedback-widget" className="footer__feedback-widget" lang={i18n.language} pageTitle={t('title')} endpoint={process.env.GATSBY_ENDPOINT} />
```
saifulislamdev, line 59: I see that the title being used here is from the translation file, but maybe you have it coming from the markdown elsewhere.

saifulislamdev (approved):

PR #39, opened 2021-07-26
Suspense

saifulislamdev:
So glad Gatsby released this version!

https://www.gatsbyjs.com/docs/reference/release-notes/v3.9/#react-18---new-suspense-ssr-architecture


package.json
```diff
@@ -30,8 +30,8 @@
     "netlify-cms-app": "^2.14.8",
     "node-sass": "^4.14.1",
     "prop-types": "^15.7.2",
-    "react": "^16.13.1",
-    "react-dom": "^16.13.1",
+    "react": "^18.0.0-alpha-419cc9c37-20210726",
```
aannyyaa, line 33: I'm not sure if we should be using an alpha version of react 18. There could be issues with using a version before general release? Can we use an older version of suspense with react 17?

aannyyaa (changes requested):

package.json
saifulislamdev: Yes, I understand what you're saying. However, older versions of `Suspense` aren't supported, not even the `Suspense` in version 17 of `React`.

aannyyaa: We could do a something like [this](https://www.gatsbyjs.com/docs/using-client-side-only-packages/#workaround-4-use-reactlazy-and-suspense-on-client-side-only)

saifulislamdev: In your example, we would still need `React.Suspense`, which won't be supported unless we upgrade.

PR #45, opened 2021-07-26
Language direction

aannyyaa:


src/components/header.jsx
```diff
@@ -55,12 +55,11 @@ function Header({ slug }) {
   );
 
   const handleClick = (langKey) => {
-    i18n.changeLanguage(langKey);
-    navigate(`/${langKey}/${slug}`);
+    i18n.changeLanguage(langKey, navigate(`/${langKey}/${slug}`));
```
saifulislamdev, line 58: Whaaaat, where'd you learn this?

src/styles/_class-overrides.scss
```diff
@@ -299,14 +299,16 @@
     //position languages at the bottom
     position: absolute;
     bottom: 0;
+    left: 0;
```
saifulislamdev, line 302: What is the order for your properties here?

src/components/header.jsx
aannyyaa: I found it [here](https://www.i18next.com/overview/api#changelanguage) when I was looking into the language direction method

src/styles/_class-overrides.scss
aannyyaa: There isn't one for now - I can implement one in my sass organization PR

src/components/header.jsx
saifulislamdev, line 58: Have you tried changing the language to see if this works?

aannyyaa: yes

saifulislamdev (approved):

PR #48, opened 2021-07-27
Refactoring routing

saifulislamdev:
Through this PR, routing can be handled entirely in the Layout component without repeating the logic in every component. However, each page and template must now incorporate the Layout component. Otherwise, the routing would not be handled on that page or template.


gatsby-node.js
```diff
@@ -4,18 +4,26 @@ const path = require('path');
 const fs = require('fs-extra');
 const { createFilePath } = require('gatsby-source-filesystem');
 const _ = require('lodash');
+// const { useTranslation } = require('react-i18next');
```
aannyyaa, line 7: What are the commented out additions in this file

saifulislamdev: Some experimenting with i18n and plugins for `gatsby-node`.

PR #58, opened 2021-07-29
Documentation (cont.)

saifulislamdev:
Further changes made.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,9 +13,3 @@ In this application:
   |- PULL_REQUEST_TEMPLATE.md
 
 ```
-  ## How to run the template site on your local computer
-
-  1. Clone the repository `git clone https://github.com/nyc-cto/LPaaS2.git`
-  1. Navigate into the directory `cd LPaaS2`
-  1. Install dependencies `npm install` or `yarn`
-  1. Run the Gatsby development server `gatsby develop`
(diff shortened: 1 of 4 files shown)


src/components/footer.jsx
```diff
@@ -79,7 +79,7 @@ function Footer() {
             )}
           />
           <Grid
-            className="usa-footer__contact-links footer"
+            className="usa-footer__contact-links footer" // TODO: contact-links?
```
aannyyaa: I updated the classname here in my most recent PR. 

Also, what do you think about putting to dos in airtable instead of as comments?

README.md
```diff
@@ -12,10 +12,4 @@ In this application:
   |- LICENCE.md
   |- PULL_REQUEST_TEMPLATE.md
 
-```
-  ## How to run the template site on your local computer
```
aannyyaa, line 16: Not sure if it makes sense to delete this until we replace it with the updated instructions

src/components/footer.jsx
saifulislamdev: I put the TODOS in the Airtable but forgot that I included this here. I'll remove it.

README.md
saifulislamdev: I had it removed because it isn't updated, and someone might follow these instructions here and get confused.

PR #68, opened 2021-08-03
Even More Styling Updates

saifulislamdev:


src/components/footer.jsx
```diff
@@ -18,15 +18,15 @@ function Footer() {
 
   const primaryLinks = links.primary;
   const primaryLinksLength = primaryLinks.length;
-  const primaryLinksLabels = t('footer.primaryLinks');
+  const primaryLinksLabels = t('navigation.dropdowns').map((element) => element.simpleLinks).flat().concat(t('navigation.parentLinks'));
```
aannyyaa: Cool! I had never seen the `.flat()` method before

src/components/footer.jsx
```diff
@@ -18,15 +18,15 @@ function Footer() {
...
   const primaryLinksLabelsLength = primaryLinksLabels.length;
   const primaryLength = primaryLinksLength > primaryLinksLabelsLength
     ? primaryLinksLabelsLength
     : primaryLinksLength;
   // take shorter length if is missing link in primaryLinks or missing label in translation file
-  if (primaryLinksLength !== primaryLinksLabelsLength) {
+  if (primaryLinksLength !== primaryLinksLabelsLength) { // TODO: upto here
```
aannyyaa: What is the to do here?

src/components/link.jsx
```diff
@@ -48,18 +48,9 @@ function Link({
...
-      variant={
-        variant !== 'nav' && 'external'
-      } /* if not nav link, set it to be regular external link */
-      className={classNames(
-        {
-          'usa-link--external usa-link--nav': variant === 'nav',
-        },
-        className,
-      )}
+      variant={variant === 'nav' && 'nav'}
```
aannyyaa, line 51: Why is `&& 'nav'` needed?

src/constants/links.js
```diff
@@ -7,7 +7,7 @@ export const headerLinks = {
 };
 
 export const footerLinks = {
-  primary: ['link-one', 'link-two', 'link-three', 'link-four'],
+  primary: headerLinks.navDropDowns.flat().concat(headerLinks.parent),
```
aannyyaa: This file should be updated to use objects to store the label (from the translation files) along with the link, to make things more clear.

src/locales/ar/translation.json
```diff
@@ -5,32 +5,26 @@
     "shortformName": "[ar] NYC CTO"
   },
   "location": "[ar] NYC",
+  "navigation": {
+    "dropdowns": [
+      {
+        "label": "[ar] Nav Label 1",
```
aannyyaa: I think "label" and "simpleLinks" are confusing. Maybe it should be something like "accordionButtonLabel" and "linkLabels"?

Also maybe "parentLinks" should be changed? The label could even just be "linkLabels" as well?

src/constants/links.js
aannyyaa: Also the key names in this file should correspond to the ones in the translation.json files, so it's easy to see how they relate

aannyyaa (changes requested):

src/components/footer.jsx
saifulislamdev: Yep, I saved so many lines of code and potential recursion on my end.

src/components/footer.jsx
saifulislamdev: It was a personal note of where I last left off 🤦🏽‍♂️

src/components/link.jsx
saifulislamdev: To specify it's a navigation link. Without it, it would be `variant={variant === 'nav'}` which would translate to `variant={true}` or `variant={false}`, neither being an actual variant (i.e. `nav` or `external`).

src/constants/links.js
saifulislamdev: > This file should be updated to use objects to store the label (from the translation files) along with the link, to make things more clear.

Yes, that will be addressed in a separate PR.

saifulislamdev: > Also the key names in this file should correspond to the ones in the translation.json files, so it's easy to see how they relate

I can only think of renaming `headerLinks.navDropDowns` to something like `headerLinks.navDropDownSimpleLinks`. Where else do you see we can do this, or what were you thinking of?

aannyyaa: I think it depends on how the object is set up and if/how we rename the keys in the translation files

src/locales/ar/translation.json
saifulislamdev: For the dropdowns, how about `buttonLabel` and `linkLabels`. And for the parent links, `linkLabels` might be confusing when they're trying to debug error messages and misinterpret which`linkLabels` to fix. How about `parentLinkLabels`, instead?

aannyyaa: `buttonLabel` and `linkLabels` work for me. I agree that it could be helpful to keep the parent labels different from `linkLabels` but I'm wondering if there's another word besides parent? In our case are they really parent links?

src/constants/links.js
saifulislamdev: Sure, we can consider this in the other PR I will create for objects.

src/locales/ar/translation.json
saifulislamdev: `buttonLabel` and `linkLabels` are now used for the dropdowns. Also, I mentioned `parentLinkLabels` since those links are referred to as parent links on the [Storybook](https://trussworks.github.io/react-uswds/?path=/story/components-header--extended-header). What do you think? It's still being referred to as `parentLinks`.

aannyyaa: I'm not sure what a better term would be, but I am not sure what makes something a parent link. Does it mean linking back to the main page of a website? In the[ landing page template](https://trussworks.github.io/react-uswds/?path=/docs/page-templates-landing-page--landing-page), it just says `simple link` and then `navigation link` in the nav dropdown menus.

aannyyaa: Also now that `simpleLinks` are just `linkLabels` in the translation files, I don't think we need the word 'simple' in `headerLinks.navDropDownSimpleLinks`. Since the links could already be confusing to someone new to the repo, I think the naming should be even more one-to-one

saifulislamdev: So `headerLinks.navDropDownLinks` instead of `headerLinks.navDropDownSimpleLinks`? And for parent links, let's have `navigation.parentLinkLabels` and `headerLinks.parentLinks`? 

Although, the naming in `src/constants/links.js` seems to be repeating the word `Links` at the end of everything in `headerLinks`.

aannyyaa: Sounds good. We can update `headerLinks` and `footerLinks` to `header` and `footer` in `src/constants/links.js` then?

saifulislamdev: - [x] `headerLinks.navDropDownSimpleLinks` → `headerLinks.navDropDownLinks` 
- [x] `navigation.parentLinks` → `navigation.parentLinkLabels`
- [x] `headerLinks.parent` → `headerLinks.parentLinks`
- [x] `headerLinks` → `header`
- [x] `footerLinks` → `footer`

aannyyaa (approved):

PR #69, opened 2021-08-04
ESLint Updates

aannyyaa:


.eslintrc
```diff
@@ -7,6 +7,9 @@
     "browser": true
   },
   "rules": {
+    "react/jsx-props-no-spreading": { "custom": "ignore" },
```
saifulislamdev, line 10: What is meant by `jsx-props-no-spreading`?

package.json
```diff
@@ -29,7 +29,6 @@
     "lodash": "^4.17.21",
     "netlify-cms-app": "^2.14.8",
     "node-sass": "^4.14.1",
-    "prop-types": "^15.7.2",
```
saifulislamdev, line 32: Did you run `npm uninstall prop-types`?

.eslintrc
aannyyaa: It doesn't allow prop spreading like `{...other}` in the code below. I set the rule so that it allows prop spreading in custom components. Also here are the [docs](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-props-no-spreading.md) for more info

```
 <GatsbyLink
        className={classNames('usa-link', className)} 
        activeClassName={activeClassName}
        partiallyActive={partiallyActive}
        to={to}
        {...other}
      >
        {children}
      </GatsbyLink>
```

package.json
aannyyaa: Yes

PR #73, opened 2021-08-05
Link, Nav, and Button Styling Updates

aannyyaa:


src/styles/_header.scss
```diff
@@ -171,6 +179,9 @@
           );
           &:hover {
             border: none;
+            &::after {
+              border: 3px solid color('secondary');
+            }
```
saifulislamdev, line 184: Which lines make the page not jump on a button hover?

src/styles/_header.scss
```diff
@@ -294,46 +299,48 @@
...
-          'base-dark'
-        );
         margin: 0 units('105');
         &:hover {
           padding-top: 0;
         }
       }
     }
   }
+  // to indicate which language is in use or active, there is a bar on one side on one side of the language button
```
saifulislamdev: `on one side on one side`

Also, it's a good idea to say where the language button is (language selector).

src/styles/_header.scss
```diff
@@ -294,46 +299,48 @@
...
+  // left-to-right languages active border
+  .banner__language-selector-item--ltr.banner__language-selector-item--active {
```
saifulislamdev, line 332: Does this count as 2 classes or 1 class?

src/styles/_header.scss
aannyyaa: Setting `$border-color` and `$border-color-hover` in the `button-theme` mixins in the global styles file

src/styles/_header.scss
aannyyaa: Not sure how I typed that twice! And I added a bit more description here and in other lines related to the language buttons

src/styles/_header.scss
aannyyaa: I think 2 but they both apply

saifulislamdev: Interesting.

saifulislamdev (approved):

PR #74, opened 2021-08-05
More Comments

saifulislamdev:


.eslintrc
```diff
@@ -7,14 +7,15 @@
     "browser": true
   },
   "rules": {
-    "react/jsx-props-no-spreading": { "custom": "ignore" },
-    "react/prop-types": "off",
-    "no-nested-ternary": "off",
+    "react/jsx-props-no-spreading": 0, // disables prop spreading messages in src/components/link.jsx
```
aannyyaa, line 10: Why were  `{ "custom": "ignore" }` and `"off"` changed to 0?

.eslintrc
```diff
@@ -7,14 +7,15 @@
...
+    "react/prop-types": 0, // disables prop type messages
+    "no-nested-ternary": "off", // disables nested ternary messages
```
aannyyaa, line 12: This rule allows for nested ternaries. In general, I think phrasing these comments as "disabling messages" is slightly different

.gitignore
```diff
@@ -1,9 +1,9 @@
+# npm
 node_modules/
 
 # Gatsby
 .cache/
-
-# Gatsby + production
+# + production
```
aannyyaa, line 6: Why is it "+ production"? Also I think `public` can just be under gatsby?

src/constants/links.js
```diff
@@ -1,13 +1,13 @@
...
+    ['link-one', 'link-two'], // first dropdown
+    ['link-three', 'link-four'], // second dropdown
   ],
-  parentLinks: ['link-five'],
+  parentLinks: ['link-five'], // parent (standalone) links
 };
 
 export const footer = {
-  primary: header.navDropDownLinks.flat().concat(header.parentLinks),
+  primary: header.navDropDownLinks.flat().concat(header.parentLinks), // taken directly from header (don't make change here)
```
aannyyaa, line 10: Definitely helpful to say not to add a change here

.eslintrc
aannyyaa: Also `{ "custom": "ignore" }` only applies to custom components, so it's more specific than just allowing prop spreading in general. For instance `` would still not be allowed under this rule

saifulislamdev: I was getting errors when formatting the code using the ESLint extension with this configuration until I changed both to 0.

.eslintrc
saifulislamdev: Replaced with `"no-nested-ternary": "off", // allows nested ternaries`

.gitignore
saifulislamdev: Replaced with 
```
# Gatsby
.cache/
public/ 
```

.eslintrc
aannyyaa: What were the errors?

.eslintrc
aannyyaa: Thanks! Should the other comments be changed too?

.eslintrc
saifulislamdev: It wouldn't format the code at all. 

Error: 
```
Configuration for rule "react/jsx-props-no-spreading" is invalid:
Severity should be one of the following: 0 = off, 1 = warn, 2 = error (you passed '{ custom: "ignore" }').
```

.eslintrc
saifulislamdev: It is currently
```
"react/jsx-props-no-spreading": "off", // allows prop spreading (used in src/components/link.jsx)
"react/prop-types": "off", // disables prop type messages
"no-nested-ternary": "off", // allows nested ternaries
```

.eslintrc
aannyyaa: Weird, that syntax is in the docs. I guess we should use 0 consistently instead of "off" then? We can change `"no-nested-ternary": "off"`

aannyyaa: Just saw your other comment - does `off` work? Just not `custom: ignore`?

.eslintrc
aannyyaa: Okay I think the second one is something more along the lines of `does not require props validation in a React component definition`

.eslintrc
saifulislamdev: `"off"` and `0` work the same. However,  `{ "custom": "ignore" }` does not work. I will change all the `"off"` to be `0`.

aannyyaa: If `off` works then maybe we should use that since it's more clear than 0. We should just be consistent about which one we use

saifulislamdev: Changed all `0`s to `"off"`.

.eslintrc
saifulislamdev: Updated!

PR #75, opened 2021-08-06
Refactoring `src/constants/links.js` to use objects

saifulislamdev:


src/components/header.jsx
```diff
@@ -28,14 +28,10 @@ function Header({ slug }) {
   const onClick = () => setExpanded((prevExpanded) => !prevExpanded);
 
   /* dynamically store parent links */
-  const parentLinks = links.parentLinks;
+  const { parentLinks } = links; // links + labels
```
aannyyaa, line 31: Nice syntax change.

src/components/header.jsx
```diff
@@ -28,14 +28,10 @@ function Header({ slug }) {
...
   const parentLinksLength = parentLinks.length;
-  const parentLinksLabels = t('navigation.parentLinkLabels');
+  const parentLinksLabels = t('navigation.parentLinkLabels'); // labels
```
aannyyaa: Should the labels be coming from the links constants file instead of from the translation file?

src/constants/links.js
```diff
@@ -1,13 +1,33 @@
 export const header = {
-  navDropDownLinks: [
-    ['link-one', 'link-two'],
-    ['link-three', 'link-four'],
+  navDropDowns: [
+    [
+      {
+        label: 'navigation.dropdowns.0.linkLabels.0',
```
aannyyaa, line 6: Now will the user be expected to update the labels in this file themselves when they add or subtract nav dropdowns?

src/components/header.jsx
saifulislamdev: The labels are coming from the links constants file, but the translation file labels are used to check if the links and labels in the links constant files are consistent with the labels in the translation files.

src/constants/links.js
saifulislamdev: Exactly, which is why I thought we were hard-coding.

src/components/header.jsx
aannyyaa: Oh okay, maybe the comment should be updated then? When I see these lines and the ` // labels` comment, I think that the labels from the translation file are being used

src/components/footer.jsx
```diff
@@ -16,12 +16,12 @@ import { logoFooter } from '../images';
...
-  const primaryLinks = links.primary;
+  const primaryLinks = links.primary; // link + labels
   const primaryLinksLength = primaryLinks.length;
   const primaryLinksLabels = t('navigation.dropdowns')
-    .map((element) => element.linkLabels)
-    .flat()
-    .concat(t('navigation.parentLinkLabels'));
+    .map((element) => element.linkLabels) // get only link labels (not dropdown button labels)
+    .flat() // unnest/flatten a nested array
+    .concat(t('navigation.parentLinkLabels')); // add parent link labels
```
aannyyaa, line 25: These new comments are helpful!

src/components/header.jsx
saifulislamdev: Good point! Comment updated to 
```// labels (used for verification only; labels are taken from parentLinks)```

src/constants/links.js
aannyyaa: Yeah... trying to think of other options

saifulislamdev: We can discuss this with Rapi during the next code review.

aannyyaa (approved):

PR #76, opened 2021-08-09
Language Config

aannyyaa:


src/components/footer.jsx
```diff
@@ -18,8 +18,8 @@ function Footer() {
 
   const primaryLinks = links.primary;
   const primaryLinksLength = primaryLinks.length;
-  const primaryLinksLabels = t('navigation.dropdowns').map((element) => element.linkLabels).flat().concat(t('navigation.parentLinkLabels'));
-  const primaryLinksLabelsLength = primaryLinksLabels.length;
+  const primaryLinksLabels = t('navigation.dropdowns').map && t('navigation.dropdowns').map((element) => element.linkLabels).flat().concat(t('navigation.parentLinkLabels'));
```
saifulislamdev: The logic in this file has changed through my PRs. I would recommend merging dev into this branch and making sure what you have here works.

src/components/footer.jsx
```diff
@@ -18,8 +18,8 @@ function Footer() {
...
+  const primaryLinksLabelsLength = primaryLinksLabels?.length;
```
saifulislamdev, line 26: Interesting, I've never seen this notation before.

src/components/i18n.js
```diff
@@ -1,65 +0,0 @@
-import i18n from 'i18next';
```
saifulislamdev, line 1: I made changes in this file that are available in dev. No major changes, but having those changes in your new file would be great.

src/components/nav-drop-down.jsx
```diff
@@ -37,7 +37,7 @@ function NavDropDown() {
       const navDropDownLinks = dropDownLinks[i];
       const navDropDownLinksLength = navDropDownLinks.length;
       const navDropDownLinkLabels = dropDownLabels[i].linkLabels;
-      const navDropDownLinkLabelsLength = navDropDownLinkLabels.length;
+      const navDropDownLinkLabelsLength = navDropDownLinkLabels?.length;
```
saifulislamdev: The logic in this file has also changed through my PRs. I would recommend merging dev into this branch and making sure what you have here works.

src/i18n-config.js
```diff
@@ -0,0 +1,65 @@
+import i18n from 'i18next';
```
saifulislamdev, line 2: How come this file is in `src/` and not `src/components/`?

aannyyaa: Since it's not a component I moved it to src

src/components/i18n.js
aannyyaa: I think they're all incorporated

saifulislamdev: Even the comment?

aannyyaa: Yeah I copied and pasted the comment over when resolving the merge conflicts

PR #78, opened 2021-08-10
404 Template

saifulislamdev:


gatsby-node.js
```diff
@@ -56,16 +49,31 @@ exports.createPages = ({ actions, graphql }) => {
...
-          component: path.resolve(
-            `src/templates/${String(edge.node.frontmatter.templateKey)}.jsx`,
-          ),
-          context: {
-            id,
-            lang,
-          },
-        });
+        // Check if the page is a localized 404 (i.e. /en/404/)
+        if (edge.node.fields.slug.match(/^\/[a-z]{2}\/404\/$/)) {
```
aannyyaa, line 60: Kudos on matching for localized and non-localized 404!

src/components/banner.jsx
```diff
@@ -9,7 +9,7 @@ import LanguageSelector from './language-selector';
 function Banner({ children, slug }) {
   const { t } = useTranslation();
   return (
-    <div className="usa-section--dark">
+    <section className="usa-section--dark">
```
aannyyaa, line 16: Why change to section?

src/components/hero.jsx
```diff
@@ -18,7 +18,6 @@ function Hero({ hero }) {
                 onClick={() => {
                   navigate(hero.buttonLink);
                 }}
-                className="usa-button"
```
aannyyaa, line 27: In my PR this will be changed to usa-button--inverse, so a class name will be used. But no change needed here

src/components/layout.jsx
```diff
@@ -14,14 +15,14 @@ function Layout({ children, slug }) {
...
     const lang = path.split('/')[1]; // language
-    const route = path
-      .split('/')
-      .slice(2)
-      .filter((v) => v !== '')
-      .join('/'); // rest of the path excluding the language
-    if (path === '/') navigate(`/${i18n.language}/`); // '/' redirect
-    else if (lang === '404') navigate(`/${i18n.language}/404`); // '/404' redirect
-    else if (lang !== i18n.language) i18n.changeLanguage(lang); // change current language if there's a different language entered in path
+    if (path === '/') navigate(`/${i18n.language}/`);
```
aannyyaa, line 18: Nice!

src/templates/404-page.jsx
```diff
@@ -0,0 +1,96 @@
...
+
+  const { markdownRemark } = data;
+  const { frontmatter } = markdownRemark;
+
+  return (
+    <>
+      <Helmet title={frontmatter.pageTitle} htmlAttributes={{ lang: i18n.language }} />
+      <Layout slug={frontmatter.slug}>
+        <main>
+          <section className="usa-section">
```
aannyyaa, line 22: You can use the section component that already exists, as we did in the landing page. Or perhaps just use the markdown body for the text sections, like in the documentation page

src/markdown-pages/404/404-ar.md
```diff
@@ -0,0 +1,16 @@
+---
+lang: ar
+templateKey: 404-page
+slug: "404"
+pageTitle: "ar 404"
+error: ar Page not found
+causeOfError: ar We’re sorry, we can’t find the page you're looking for. It might have been removed, changed its name, or is otherwise unavailable.
```
aannyyaa, line 7: - As I mention in another comment, it might make sense to just use the markdown body for this text.
- When you do use the front matter, quotation marks are not needed around strings. 
- I'm not sure we need the ways to fix error section. I know it's in the template 404 page, but our URLs can follow multiple potential patterns
- Also for the visit our home page section, I'm not sure if we should include the 'helpful tools and resources part'

aannyyaa (changes requested):

src/components/banner.jsx
saifulislamdev: For consistency, you use sections in the other components.

src/components/hero.jsx
saifulislamdev: So I can leave it as it is?

aannyyaa: Yes

src/templates/404-page.jsx
saifulislamdev: I see what you mean, but the contents inside the section component don't match the content for the 404 pages. For example, we would need multiple text and buttons, while the section component only has placeholders for one text and one button. We could use multiple sections for the 404 page to accomplish this, but it wouldn't make sense to use multiple sections when the 404 content is all related. 

Also, there is an `h1` tag on the 404 pages, but the section component does not have one. We can totally use markdown body for the text sections, but by doing it the current way, we're making the 404 pages more easily editable.

src/markdown-pages/404/404-ar.md
saifulislamdev: * Addressed in the other comment, can ignore that here, unless a change is made.
* In this case, quotation marks are needed because there is a type error. For example, I cannot have `home` and `404` for slugs, as they are different types (I was getting an error when I did it like this). After making `404` a string, the error was resolved.
* It might be an accessibility thing that we should take into account. We can keep it like this for now, and discuss it during code review?
* We should also discuss this during code review.

src/templates/404-page.jsx
aannyyaa: I think using markdown body would be more easily editable, since you can clearly see the text layout

src/markdown-pages/404/404-ar.md
aannyyaa: Oh okay the type error makes sense. Maybe the slug should be `not-found` and the page title can be "Page Not Found" or something along those lines?

src/templates/404-page.jsx
saifulislamdev: Markdown body would make sense for something like the documentation where it's all text. However, in cases like this where code is also incorporated, it would be best to go about it the way we do now. Otherwise, we would have used the markdown body for the landing pages to start with.

src/markdown-pages/404/404-ar.md
saifulislamdev: Gatsby tends to use `404` as the slug, as we saw in my code in `gatsby-node.js`, which was taken from Gatsby documentation. The page title makes sense, though; I'll set it to be `404 | Page Not Found`.
```javascript
// Check if the page is a localized 404 (i.e. /en/404/)
  if (edge.node.fields.slug.match(/^\/[a-z]{2}\/404\/$/)) {
```

aannyyaa: Okay sounds good

src/templates/404-page.jsx
aannyyaa: I was thinking you could use markdown for everything but the buttons, but we can leave as is

PR #83, opened 2021-08-12
i18n debug

saifulislamdev:


src/pages/404.js
```diff
@@ -5,7 +5,7 @@ import Layout from '../components/layout';
 
 function NotFound() {
   return (
-    <Layout />
+    <Layout languageList={[]} />
```
aannyyaa: Is there any content shown on the 404 page? There is text for the 404 message in the translation files

aannyyaa: Never mind - about to review the 404 PR!

PR #84, opened 2021-08-12
Constant Code Convention in `src/constants/links.js`

saifulislamdev:


aannyyaa (approved):

src/constants/links.js
```diff
@@ -3,38 +3,44 @@ export const header = {
...
+
+
 export const footer = {
   primary: header.navDropDowns
     .flat() // turn nested array into flat/unnested array
     .concat(header.parentLinks), // add parent links (taken directly from header; don't make change here)
   secondary: {
-    terms: 'terms',
-    privacy: 'privacy',
+    TERMS: 'terms',
```
aannyyaa, line 42: Should these also be objects with link and label?

saifulislamdev: It's not as difficult to find the labels for the term and privacy compared to the dropdowns in the translation files, so this is fine.

PR #89, opened 2021-08-16
Update Links

aannyyaa:


src/constants/links.js
```diff
@@ -25,16 +26,19 @@ export const header = {
     ],
     // second dropdown end
   ],
-  parentLinks: [{ LABEL: 'navigation.parentLinkLabels.0', LINK: 'link-five' }], // parent (standalone) links
+  parentLinks: [
+    { LABEL: 'navigation.parentLinkLabels.0', LINK: 'home' },
+    { LABEL: 'navigation.parentLinkLabels.1', LINK: 'documentation' },
```
aannyyaa, line 31: Do we want "Home" and "Documentation" to be parent links or nav dropdown links? I'm leaning toward parent links because the user is more likely to see them that way. 

Also should we make it so that clicking on the MOCTO logo takes people back to the home page?

src/styles/_custom-theme.scss
```diff
@@ -74,7 +81,7 @@ $theme-color-secondary-darker: $custom-secondary-darker;
 
 // currently, accent cool is the same as primary
 //if you have the need for accent color palettes, you can set them here
-$theme-color-accent-cool-lighter: white; //set to white in order to access white in mixins
```
aannyyaa, line 77: Turns out I can access white in mixins without setting it as one of the theme colors, I guess it is its own color token

src/constants/links.js
```diff
@@ -1,14 +1,15 @@
 export const header = {
-  navDropDowns: [ // dropdowns
+  navDropDowns: [
+    // dropdowns
     // first dropdown start
     [
       {
         LABEL: 'navigation.dropdowns.0.linkLabels.0',
-        LINK: 'link-one',
+        LINK: 'home',
```
saifulislamdev: `/home`?

src/constants/links.js
```diff
@@ -1,14 +1,15 @@
...
       },
       {
         LABEL: 'navigation.dropdowns.0.linkLabels.1',
-        LINK: 'link-two',
+        LINK: 'documentation',
```
saifulislamdev: `/documentation`?

src/constants/links.js
saifulislamdev: Parent links, they aren't related. Great idea; clicking the MOCTO logo can take people back to the home page. You can add it to the backlog or go about it in this PR if you'd like.

src/locales/en/translation.json
```diff
@@ -9,14 +9,14 @@
     "dropdowns": [
       {
         "buttonLabel": "Nav Label 1",
-        "linkLabels": ["Simple link one", "Simple link two"]
+        "linkLabels": ["Home", "Documentation"]
```
saifulislamdev: Reminder to remove from here.

src/styles/_footer.scss
```diff
@@ -107,8 +116,8 @@
     padding: units(3) 0;
   }
 
-  .usa-footer__primary-link {
-    padding: units(1.5);
+  .usa-link.usa-footer__primary-link {
+    padding: units(1.5) 0 0 0;
```
saifulislamdev: Are you trying to override padding here?

src/styles/_header.scss
```diff
@@ -55,14 +55,16 @@
...
-          &:hover {
-            border-left: units(0.5) solid color('secondary');
-            padding-left: units(0.5);
-          }
+          // &:hover {
+          //   p {
+          //     border-left: units(0.5) solid color('secondary');
+          //     padding-left: units(0.5);
+          //   }
+          // }
```
saifulislamdev, line 67: Do you want to keep it as comments?

src/styles/_footer.scss
aannyyaa: What do you mean by override? There's bigger padding in mobile/tablet view, so the tap area is bigger. I guess I could have more bottom padding in desktop as well. I'd just have to reposition the underline

src/styles/_header.scss
aannyyaa: It's a placeholder for now

src/constants/links.js
aannyyaa: Okay great - I'll add it to this PR

src/constants/links.js
aannyyaa: for some reason `home` and `/` both work but `/home` doesn't

src/constants/links.js
aannyyaa: Doesn't route correctly when the slash is included

src/locales/en/translation.json
aannyyaa: Done

src/constants/links.js
aannyyaa: It must have to do with the logic in our custom link component?

src/locales/es/translation.json
```diff
@@ -16,7 +16,7 @@
         "linkLabels": ["Enlace Simple tres", "Enlace Simple cuatro"]
       }
     ],
-    "parentLinkLabels": ["Parent link"]
+    "parentLinkLabels": ["Página de Inicio", "Documentación Técnica"]
```
aannyyaa, line 19: Let's double check that these are the correct phrases

src/styles/_footer.scss
aannyyaa: updated

src/constants/links.js
saifulislamdev: I completely forgot about the language code in the pathname; this is fine how it is. I'm not sure if Reach Router can help us or if it's beyond the scope of Reach Router? 

Can having `<Router basepath={i18n.language} />` help us in this case?

src/styles/_footer.scss
saifulislamdev: By override, I meant that there is a default padding, and you're overriding it by adding 0s. This is fine.

src/constants/links.js
aannyyaa: I'll look into it!

src/locales/es/translation.json
saifulislamdev: Verified with Oscar.

src/styles/_footer.scss
aannyyaa: Okay - yeah it's not overriding the default - it's just changing between the mobile and desktop views

src/constants/links.js
aannyyaa: I think if we wanted to use a base path we could use [withPrefex](https://www.gatsbyjs.com/docs/reference/built-in-components/gatsby-link/#add-the-path-prefix-to-paths-using-withprefix) from gatsby

saifulislamdev: We could use `withPrefix` if the prefix were static. However, our prefix is dynamic. If you take a look [here](https://www.gatsbyjs.com/docs/how-to/previews-deploys-hosting/path-prefix/), the prefix is done in `gatsby-config`, where i18n is not available. I think we can have it how it is right now.

saifulislamdev (approved):

PR #92, opened 2021-08-19
Final Cleanup

saifulislamdev:


src/pages/404.js
```diff
@@ -4,22 +4,13 @@
...
 
-import React, { useEffect } from 'react';
-
-import { navigate } from 'gatsby';
-import { useTranslation } from 'react-i18next';
+import React from 'react';
 
 import Layout from '../components/layout';
-
+import i18next from '../i18n-config'; // do not remove (affects routing)
```
aannyyaa, line 10: Where is i18next used in the file now?

saifulislamdev: i18next is used in the layout.

